PR #24, opened 2012-06-19
Add .travis.yml for Travis CI (http://travis-ci.org/)

msabramo:
To set up Travis for your repo, see http://about.travis-ci.org/docs/user/getting-started/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+language: java


michaelklishin:
We do provide Cassandra (that needs to be started in a `before_install` step) on travis-ci.org but HBase is not (officially) provided and we may or may not have the version Titan needs. In addition, the wiki says

```
Note, that running the comprehensive test suite requires a significant amount of of time (> 1 hour)
```

This is too long for travis. Builds that take longer than 25-30 minutes will be killed (with ~14K projects on a free service, I hope this sounds reasonable).

That said, if there is a way to run a subset of tests for Titan and it will be beneficial enough, we are happy to see Titan on travis. While you are at it, consider [testing against multiple JDKs](http://about.travis-ci.org/blog/support_for_multiple_jdks/)

Starting cassandra should be as easy as adding

``` yaml
before_install:
  - sudo service cassandra start
  - sleep 3
  - sudo service cassandra status
```

MK on behalf of the travis-ci.org team.

okram:
http://travis-ci.org/#!/thinkaurelius/titan/
  http://travis-ci.org/thinkaurelius/titan/

I will look into @michaelklishin notes on the cassandra service call. Good stuff. Thanks.

Issue #26, opened 2012-06-20
NoClassDefFoundError in bin/gremlin.sh


jncorpron:
Running bin/gremlin.sh in cygwin causes a NoClassDefFoundError:

java.lang.NoClassDefFoundError: com/thinkaurelius/titan/tinkerpop/gremlin/Console
Caused by: java.lang.ClassNotFoundException: com.thinkaurelius.titan.tinkerpop.gremlin.Console
        at java.net.URLClassLoader$1.run(URLClassLoader.java:202)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:190)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:306)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:301)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
Could not find the main class: com.thinkaurelius.titan.tinkerpop.gremlin.Console.  Program will exit.
Exception in thread "main"

This is apparently caused by using the wrong path separator ":" for windows.

PR #27, opened 2012-06-20
Fix for #26

jncorpron:
I'm using case here since it was used in another script I have that can be expanded to detect other OSs and handle other peculiarities. It may also be better to extract the file separation character into a variable so that there is only one "CP=..." line.

### DIFF
--- a/src/main/bin/gremlin.sh
+++ b/src/main/bin/gremlin.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-case `uname -o` in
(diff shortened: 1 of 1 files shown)


okram:
When I merge, I get this issue on MacOSX.

gremlin$ ./gremlin.sh 
uname: illegal option -- o
usage: uname [-amnprsv]

```
     \,,,/
     (o o)
```

-----oOOo-(_)-oOOo-----

PR #42, opened 2012-07-30
[HBase] limiting amount of columns with filter

crisweber:
Hi there!

[07-30]
I worked a bit on titan this weekend, and I started changing getSlice to apply column limiting at server side. I used ColumnPaginatorFilter, which was created after ColumnCountFilter to be kind of SQL's LIMIT/OFFSET.

[08-03]
Found that clearStorage method was not using the same Configuration used in openDatabase method. I was testing with a different Zookeeper port when I found it.

### DIFF
--- a/src/main/java/com/thinkaurelius/titan/diskstorage/hbase/HBaseOrderedKeyColumnValueStore.java
+++ b/src/main/java/com/thinkaurelius/titan/diskstorage/hbase/HBaseOrderedKeyColumnValueStore.java
@@ -17,8 +17,10 @@ import org.apache.hadoop.hbase.client.HTablePool;
 import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.Row;
+import org.apache.hadoop.hbase.filter.ColumnPaginationFilter;
 import org.apache.hadoop.hbase.filter.ColumnRangeFilter;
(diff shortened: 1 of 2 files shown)


dalaro:
This is great, thanks.  b77cfe3 is a good catch.

The HBase tests passed on my box with this applied under HBase 0.92.1.

I wonder if 476cde1 would take a while on large tables, but since clearStorage is used primarily in testing on small tables, I don't think it's a material concern.

crisweber:
Hi Dan!

Yes, probably 476cde1 will take a while on large tables. If clearStorage became a feature it will be better to keep a method for large tables dropping it and recreating and a subclass with this method overwritten by the 'remove it all' impl only for test purposes. Maybe this subclass could be a specific StorageConfiguration. Having two different implementations for this method will require different test cases also.

I can do this, if you think that's a good approach!

Thanks,
Cristofer

dalaro:
Hi Cristofer,

I think the approach in your patch is fine.  It might be neat to have a single clearStorage() method that first uses your approach while timing itself, and eventually "gives up" and drops the table if it takes too long.  But clearStorage() is really only used in testing right now with small tables, so your approach should win in those cases.

I pulled this into master initially, but I really should have put it in titan0.1, since master commits should essentially be limited to bugfixes.  I just cherry-picked it over to titan0.1 and reverted on the master for now.  This is still slated for the next release.

thanks,
Dan

PR #56, opened 2012-08-16
Vertexes with unique key properties aren't returned by getVertices

DanielThomas:
Oversight in this method - keyIndex isn't checked, though if the transaction is committed, they'll be retrieved from disk successfully.

### DIFF
--- a/src/main/java/com/thinkaurelius/titan/graphdb/transaction/AbstractTitanTx.java
+++ b/src/main/java/com/thinkaurelius/titan/graphdb/transaction/AbstractTitanTx.java
@@ -467,8 +467,9 @@ public abstract class AbstractTitanTx extends TitanBlueprintsTransaction impleme
                 vertices.addAll(attrSubindex.get(attribute));
             }
             Map<Object, TitanVertex> keySubindex = keyIndex.get(key);
-            if (keySubindex != null) {
-                vertices.add(keySubindex.get(attribute));
+            TitanVertex vertex = keySubindex.get(attribute);
+            if (keySubindex != null && vertex != null) {
+                vertices.add(vertex);
             }
             return vertices;
         } else {
--- a/src/main/java/com/thinkaurelius/titan/graphdb/transaction/AbstractTitanTx.java
+++ b/src/main/java/com/thinkaurelius/titan/graphdb/transaction/AbstractTitanTx.java
@@ -8,7 +8,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Multimap;
 import com.thinkaurelius.titan.core.*;
-import com.thinkaurelius.titan.core.InvalidElementException;
 import com.thinkaurelius.titan.graphdb.blueprints.TitanBlueprintsTransaction;
 import com.thinkaurelius.titan.graphdb.database.InternalTitanGraph;
 import com.thinkaurelius.titan.graphdb.query.ComplexTitanQuery;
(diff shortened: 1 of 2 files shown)


DanielThomas:
This isn't quite ready, I found a problem with it...

DanielThomas:
Good to go.

DanielThomas:
Actually, the null problems I ran into raises a red flag: there's an awful lot of null state in this class. I should have some time to look at that separately.

mbroecheler:
what do you mean by "null state"?

On Thu, Aug 16, 2012 at 12:23 PM, Danny Thomas notifications@github.comwrote:

> Actually, the null problems I ran into raises a red flag: there's an awful
> lot of null state in this class. I should have some time to look at that
> separately.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/56#issuecomment-7796277.

## 

Matthias Broecheler, PhD
http://www.matthiasb.com
E-Mail: me@matthiasb.com

DanielThomas:
Collection instances that are by default null, instead of empty static instances, and later initialised when needed, gets from collections with == null, rather than using contains\* and null adverse collections would be a useful safety net. Null has its place, but in these cases it makes for more reliable, less ambigous and better to read code.

Of course, the assertion and preconditions checks already in place help and I won't touch anything unless unit test coverage is solid (and I'm sure it is - the tests are very comprehensive!). I figured because this is a class that I've worked on several times already, it's an ideal place for this kind of improvement. What do you think?

mbroecheler:
Ah, I see what you mean. For example, using newNodes=null instead of an
empty static immutable set NO_NODES as in this exemplary commit:
https://github.com/thinkaurelius/titan/blob/titan0.1/src/main/java/com/thinkaurelius/titan/graphdb/transaction/AbstractTitanTx.java

Yes, I agree - static empty containers are better than null and the code is
inconsistent in that regard. But how would that help for the attribute and
keyIndex maps? Would you change:

```
        Multimap<Object, TitanVertex> attrSubindex =
```

attributeIndex.get(key);
            if (attrSubindex != null) {
                vertices.addAll(attrSubindex.get(attribute));
            }

to

```
        if (attributeIndex.contains(key)) {
            vertices.addAll(attributeIndex.get(key).get(attribute));
        }
```

The reason I did the former is to avoid the contains method call, but I
guess that thoroughly qualifies as premature optimization ;-)
So, yeah, feel free to clean it up or file an issue for me.

On Thu, Aug 16, 2012 at 1:44 PM, Danny Thomas notifications@github.comwrote:

> Collection instances that are by default null, instead of empty static
> instances, and later initialised when needed, gets from collections with ==
> null, rather than using contains\* and null adverse collections would be a
> useful safety net. Null has its place, but in these cases it makes for more
> reliable, less ambigous and better to read code.
> 
> Of course, the assertion and preconditions checks already in place help
> and I won't touch anything unless unit test coverage is solid (and I'm sure
> it is - the tests are very comprehensive!). I figured because this is a
> class that I've worked on several times already, it's an ideal place for
> this kind of improvement. What do you think?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/56#issuecomment-7798755.

## 

Matthias Broecheler, PhD
http://www.matthiasb.com
E-Mail: me@matthiasb.com

DanielThomas:
Precisely, but I don't think need to go that far: Collections.empty\* will do it with isEmpty rather than comparing to a default instance. Anyway, I'll take a run at it :)

mbroecheler:
Cool, thanks

On Thu, Aug 16, 2012 at 3:31 PM, Danny Thomas notifications@github.comwrote:

> Precisely, but I don't think need to go that far: Collections.empty\* will
> do it with isEmpty rather than comparing to a default instance. Anyway,
> I'll take a run at it :)
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/56#issuecomment-7801741.

## 

Matthias Broecheler, PhD
http://www.matthiasb.com
E-Mail: me@matthiasb.com

PR #57, opened 2012-08-17
AbstractTitanTx null ambiguity cleanup

DanielThomas:
I ran into NPEs when I fixed a bug in this class due to some null ambiguity issues. This pull requests removes the use of null where it's not part of a contracted method return:
- Use Guava Optional for newNodes
- Collection.get\* == null has two possible meanings - the key doesn't exist or the key has a null value, use contains\* instead (maybe consider null adverse collections which don't allow nulls in keys or values)
- Set a few fields to final and removed unneeded null dereferencing on close - clear() is called on those collections anyway


DanielThomas:
I also cleaned up the mixed tab/space indentation and trailing whitespace, but in a separate commit so you don't have to navigate those changes to see the actual work I did.

mbroecheler:
Merged the pull request manually. Accepted all changes but since I changed the VertexCache contract to not allow null vertices I left the getExistingVertex method to using node!=null for slightly better performance since this method can be called very frequently.

PR #59, opened 2012-08-21
Static analysis plugins for Maven site generation

DanielThomas:
This adds Findbugs and PMD plugins to the Maven reporting configuration. Run 'mvn site' to generate the reports.

I personally prefer Sonar for performing static analysis - but this is a good starting point. There are many PMD rulesets not enabled that could be beneficial, that Sonar will run by default:

http://pmd.sourceforge.net/rules/index.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -457,8 +457,13 @@
             </configuration>
           </plugin>
           <plugin>
-            <artifactId>maven-pmd-plugin</artifactId>
-            <version>2.7.1</version>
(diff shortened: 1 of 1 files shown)


dalaro:
The FindBugs output looks promising.  Thanks for configuring this.

jramsdale:
This uses the Maven 2 style of reporting configuration but Maven 3 has been out for quite some time. Shouldn't we go with the newer syntax?

okram:
Yes. We should be using Maven3.

Marko.

PR #61, opened 2012-08-23
Refactor Titan into multi-module Maven build

jramsdale:
Posting this pull request for discussion (a la GitHub itself: https://github.com/blog/1124-how-we-use-pull-requests-to-build-github). It is NOT ready to merge. Here are the things I changed:
- Converted root from jar to pom packaging; added submodules for titan-core, titan-test, titan-distribution, and each StorageManager type
- Changed groupId from com.thinkaurelius to com.thinkaurelius.titan
- Removed redundant info in root pom (duplicates Sonatype pom or configures to defaults)
- Moved assembly configuration to titan-distribution to avoid chicken-egg build problem (standard Maven practice)
- Created new com.thinkaurelius.titan.diskstorage.util.ByteBufferUtil.bytesToHex(byteBuffer) (and corresponding unit test) to avoid Cassandra dependency from non-Cassandra modules. It involves an array copy, so could perhaps be improved upon, but it's only used in toString() methods.
- Changed code in one or two places where dependency ordering resulted in different artifacts with API changes. There is Titan technical debt here--old or conflicting dependencies should be updated and resolved whenever possible, but I wanted to defer to a later commit so as not to confuse this commit's concerns.
- Added SPI-based StorageManagerFactory since GraphDatabaseConfiguration can't know what StorageManager implementations are available at compile time (it's a dependency of implementation modules).
- Removed unnecessary uber-jar assembly

Here's what remains:
- Assemblies are not building with 'provided' scope dependencies
- Distribution assembly doesn't build with src and docs

I would appreciate any feedback. Additionally, I attempted a rebase with the titan0.1 branch and there are some issues that need to be resolved. I'll initiate an e-mail on the list concerning branching strategies--I'm not sure I understand the intent of branching prior to release.


jramsdale:
Pull request has been rebased to titan0.1. New pull request will be issued.

PR #63, opened 2012-08-24
Cleaned up several classes as part of a coding standards discussion

bobbyno:
- Whitespace for many is a highly personal thing, but the very compact style in the code now seems much more reminiscent of C++ than Java. Consistent whitespace sets a standard that many contributors will appreciate.
  *\* _id > NoID is easier to read than _id>NoID
  *\* Have a look at the changes in the com.thinkaurelius.titan.graphdb.adjacencylist package in this pull request to see the difference. I use IntelliJ to apply code formatting rules automatically; if you like this style, let me know and I'll make the format consistent. 
- else after an if statement that returns a response is redundant and adds a deeper level of nesting, making the code that much harder to read. See BasicElement and ArrayAdjacencyList for examples.
- if statements used as guard clauses don't need braces.
- if / else blocks should have braces for readability.
- removed unused imports.
- Fixed a redundant cast in ArrayAdjListFactory
- Remove commented code. It gets out of date and wastes the time of people who stop to read it to figure out what it _might_ be good for. Store experiments on a feature branch.

### DIFF
--- a/src/main/java/com/thinkaurelius/titan/core/TitanGraph.java
+++ b/src/main/java/com/thinkaurelius/titan/core/TitanGraph.java
@@ -51,6 +51,4 @@ public interface TitanGraph extends Graph, KeyIndexableGraph, ThreadedTransactio
      * @see #shutdown()
      */
     public boolean isOpen();
-
-	
 }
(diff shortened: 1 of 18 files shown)


mbroecheler:
Will do this in one go against the entire code base. Thanks for the example, Bobby.

PR #65, opened 2012-08-27
Refactor Titan into multi-module Maven build

jramsdale:
Posting this pull request for discussion (a la GitHub itself: https://github.com/blog/1124-how-we-use-pull-requests-to-build-github). It is not quite ready to merge. Here are the things I changed:
- Converted root from jar to pom packaging; added submodules for titan-core, titan-test, titan-distribution, and each StorageManager type
- Changed groupId from com.thinkaurelius to com.thinkaurelius.titan
- Removed redundant info in root pom (duplicates Sonatype pom or configures to defaults)
- Moved assembly configuration to titan-distribution to avoid chicken-egg build problem (standard Maven practice)
- Created new com.thinkaurelius.titan.diskstorage.util.ByteBufferUtil.bytesToHex(byteBuffer) (and corresponding unit test) to avoid Cassandra dependency from non-Cassandra modules. It involves an array copy, so could perhaps be improved upon, but it's only used in toString() methods.
- Changed code in one or two places where dependency ordering resulted in different artifacts with API changes. There is Titan technical debt here--old or conflicting dependencies should be updated and resolved whenever possible, but I wanted to defer to a later commit so as not to confuse this commit's concerns.
- Added SPI-based StorageManagerFactory since GraphDatabaseConfiguration can't know what StorageManager implementations are available at compile time (it's a dependency of implementation modules).
- Removed unnecessary uber-jar assembly

Here's what remains:
- Assemblies are not building with 'provided' scope dependencies
- Distribution assembly doesn't build with src and docs

I would appreciate any feedback as well as assistance with the final 2 issues.


dalaro:
Thanks.  This looks thoughtful and thorough.  I'll need some time to review.

dalaro:
I am still sorting through this.  I'm interested in the objective of letting users choose which parts of Titan they want and avoiding the rest, but this may be an item for the next release.

Regarding assemblies without provided artifacts.  I thought that was the point of provided: for depending on artifacts that are part of the expected runtime environment and therefore safe to exclude from assemblies.  What am I missing?

jramsdale:
Hi Dan,

I think the restructuring better represents the functional separation of
the code. It also provides a model for extension if others would like to
explore alternative data stores to back Titan. These already existed in the
code, but converting to modules makes this more clear and enforces certain
separation of concerns (via compile-time dependencies).

You're correct that 'provided' dependencies should exist in a runtime
environment, but Titan doesn't run in a container, so there IS no provided
runtime environment. By building our own assemblies we need full control of
the runtime classpath. This was the case even before the restructuring, but
there is something about how the Maven Assembly Plugin functions that
included 'provided' dependencies prior but isn't working now. I believe it
may have something to do with the introduction of ModuleSets and it's
possible there's a Maven bug here, but I haven't been able to figure out
how to include all the dependencies we require in the assembly. Everything
seems to be there except the 'provided' dependencies.

-jeff

okram:
Hi,

Also, Central Repo won't take a distribution larger than 32 megs. This is what has plagued Rexster for the last 3 released until we went multi-module in 2.1.

Marko.

http://markorodriguez.com

Issue #66, opened 2012-08-28
Should storage read operations should be avoided in StandardTitanGraph init?


DanielThomas:
I've configured a Cassandra cluster and can't get Titan to start with an embedded Cassandra instance because I can't get a quorum (TitanFactory.open is on a critical path that will cause the service to fail if it fails).

I'll workaround by retrying the graph database open, but shouldn't storage dependencies be avoided in constructors, instead calling StandardIDPool.renewBuffer() lazily? That would allow a Titan instance to be returned, and individual graph operations to fail due to any underlying storage problem, rather than throwing an exception:

```
com.thinkaurelius.titan.core.GraphStorageException: Exception in storage backend.
        at com.thinkaurelius.titan.diskstorage.cassandra.CassandraThriftOrderedKeyColumnValueStore.getSlice(CassandraThriftOrderedKeyColumnValueStore.java:180)
        at com.thinkaurelius.titan.diskstorage.cassandra.CassandraThriftOrderedKeyColumnValueStore.getSlice(CassandraThriftOrderedKeyColumnValueStore.java:192)
        at com.thinkaurelius.titan.diskstorage.util.OrderedKeyColumnValueIDManager.getIDBlock(OrderedKeyColumnValueIDManager.java:68)
        at com.thinkaurelius.titan.diskstorage.cassandra.CassandraThriftStorageManager.getIDBlock(CassandraThriftStorageManager.java:195)
        at com.thinkaurelius.titan.graphdb.database.idassigner.StandardIDPool.renewBuffer(StandardIDPool.java:73)
        at com.thinkaurelius.titan.graphdb.database.idassigner.StandardIDPool.<init>(StandardIDPool.java:42)
        at com.thinkaurelius.titan.graphdb.database.idassigner.SimpleVertexIDAssigner.<init>(SimpleVertexIDAssigner.java:42)
        at com.thinkaurelius.titan.graphdb.configuration.GraphDatabaseConfiguration.getIDAssigner(GraphDatabaseConfiguration.java:312)
        at com.thinkaurelius.titan.graphdb.database.StandardTitanGraph.<init>(StandardTitanGraph.java:83)
        at com.thinkaurelius.titan.core.TitanFactory.open(TitanFactory.java:60)
Caused by: UnavailableException()
        at org.apache.cassandra.thrift.Cassandra$get_slice_result.read(Cassandra.java:7288)
        at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:78)
[...]

PR #67, opened 2012-08-28
Quick fix to lazy initialise ID pools

DanielThomas:
This isn't necessarily the right way to do this, but I needed graph constructors to have no storage dependencies so lack of Cassandra quorum on startup doesn't lead to a chicken-egg situation. Probably needs more discussion. Issue #66.


mbroecheler:
This has been incorporated in a recent commit.

PR #77, opened 2012-09-04
Fix stale lock bug in LockingTransactionHandle

tedoc2000:
This fixes a bug where locks can be left in a cassandra db when the lock creation process is too slow.


dalaro:
Thanks.  The patch looks correct.  Changing the timestamp part of a lock column when going to delete it is obviously wrong.  I probably conflated the timestamp on the write command (which is handled at a lower level) with the timestamp portion of the lock column when I wrote this bug.  This bug would have caused locks written in longer than lockWaitMS to stay in Cassandra, causing false contention up to the lock expiration time.

The Travis failure is due to some process on the build machine holding TCP port 7000.  The Cassandra daemon in our test suite wants that port.  The failure looks unrelated to the patch.

I don't understand your report sentence.  I thought this bug caused locks to be left in Cassandra when lock writing is too slow, rather than too fast as you wrote.  Would you mind commenting?

tedoc2000:
Whoops sorry I misread the inequality.. Let me fix the commit message.

PR #201, opened 2013-03-31
Use containsKey not contains

mebigfatguy:
Code uses contains as a guard against doing a put, but contains checks values, not keys...so that test will always succeed (as contains will always return false) ..  use containsKey instead.


mbroecheler:
That's a great catch - thank you. The in-memory store has undergone significant change to increase performance so I fixed it in the code rather than merge the pull.

PR #238, opened 2013-04-25
report an error when ordered partitioner is used for g.{V, E} with Astyanax

xedin:


xedin:
Also because that doesn't work anyway as the have RandomPartitioner hardcoded there).

Issue #268, opened 2013-05-10
Add Metrics monitoring


dalaro:
We would like to instrument Titan using Metrics.   Let's consider what to measure with Metrics and what application(s) we plan to initially use atop Metrics for charting (e.g. ganglia or graphite).

espeed:
From a previous email...

Here are the three leading application performance management systems (in order) -- see http://www.compuware.com/application-performance-management/compuware-leadership-in-the-APM-market.html. 

dynaTrace and AppDynamics specifically support Cassandra, and AppDynamics has partnered with DataStax (see the Martin Stone video below). 

Opnet
- http://www.opnet.com/solutions/application_performance/

dynaTrace
- http://www.compuware.com/application-performance-management/
- http://www.compuware.com/application-performance-management/dynatrace-enterprise.html
- http://apmblog.compuware.com/2011/09/20/cassandra-write-performance-a-quick-look-inside/
- https://apmcommunity.compuware.com/community/display/PUB/Analyze+and+Monitor+Big+Data+Solutions
- http://www.compuware.com/application-performance-management/dynatrace-faq.html

AppDynamics (used by Netflix, DataStax partner)
- http://www.appdynamics.com/products/why-appdynamics
- Martin Stone - Deep Visibility into Apache Cassandra with AppDynamics (http://www.youtube.com/watch?v=VZHTmJT7fvk)
- http://www.appdynamics.com/press-release/appdynamics-partners-datastax-manage-performance-and-availability-business-critical
- http://www.appdynamics.com/blog/tag/cassandra/

---//---

Killing two birds with one stone...

There is also Twitter's Zipkin monitoring system, which hooks into Finagle, Twitter's open-source RPC system for the JVM used to construct high-performance and high-concurrency servers:
- https://github.com/twitter/zipkin/
- https://github.com/twitter/finagle
- https://github.com/twitter/finagle/tree/master/finagle-zipkin
- http://engineering.twitter.com/2013/01/braindump.html

Both are Apache 2.0 licensed. 

Finagle might be a good candidate for Rexster 3.0 or TitanServer. It's built on Netty and implements uniform client and server APIs for several protocols. Most of Finagle’s code is protocol agnostic, simplifying the implementation of new protocols. 

See the number of protocols it already supports: https://github.com/twitter/finagle

Matt Ho gives a great 35 min presentation on Finagle and the considerations/challenges to think about when designing an RPC system/protocol:

[...]

PR #285, opened 2013-05-22
Added support for Rexster 2.4.0 Metrics

zachkinstner:
This updates Titan to use TinkerPop 2.4.0-SNAPSHOT.  

Support for Metrics in Titan is dependent upon these changes in Rexster 2.4.0: https://github.com/tinkerpop/rexster/pull/307

I was able to get the metrics to work by making `ReporterConfig` setup occur first -- I had to move `RexProRexsterServer` and `HttpRexsterServer` instantiation out of the `RextsterTitanServer` constructor and into the `start` method instead.  It seems to work fine, and I suppose it shouldn't cause an issue as long as `start` is only called once.

There was one issue in `TitanSpecificBlueprintsTestSuite` that I didn't know how to fix. This particular test should fail every time.

More discussion about Titan Metrics: https://github.com/thinkaurelius/titan/issues/268

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ target/
 titan-hbase/logs
 titan-hbase/src/test/titan-hbase-test-data/*
 titan-hbase/src/test/titan-zookeeper-test/*
+.classpath
+.project
+*.prefs
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
     <properties>
         <titan.version>0.3.1</titan.version>
         <titan.compatible.versions>0.3.1,0.3.0</titan.compatible.versions>
-        <tinkerpop.version>2.3.0</tinkerpop.version>
+        <tinkerpop.version>2.4.0-SNAPSHOT</tinkerpop.version>
         <junit.version>4.8.1</junit.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
--- a/titan-all/pom.xml
+++ b/titan-all/pom.xml
@@ -31,6 +31,11 @@
             <artifactId>titan-es</artifactId>
             <version>${titan.version}</version>
         </dependency>
+        <dependency>
+			<groupId>com.yammer.metrics</groupId>
+			<artifactId>metrics-core</artifactId>
+			<version>3.0.0-BETA1</version>
+		</dependency>
     </dependencies>
     <build>
         <directory>${basedir}/target</directory>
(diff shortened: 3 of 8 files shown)


dalaro:
Thanks!

Can you help me understand your changes to RexsterTitanServer?  I think you mentioned that these changes were necessary in #268.  I seem to have nearly identical changes already locally committed except for your RexsterTitanServer stuff (and the eclipse ignores).

Sorry about the duplicated work.  I postponed pushing until resolving a test failure in Titan-BDB that appeared with Blueprints 2.4.0-SNAPSHOT, but I think that's now fixed with https://github.com/tinkerpop/blueprints/pull/402.

I fixed (?) TitanSpecificBlueprintsTestSuite by just passing the String "friend" as the final argument to addEdge.  At least, I think I fixed it because the test passes this way.

zachkinstner:
> Sorry about the duplicated work.

No worries -- I'm just trying to help out where I can.

> Can you help me understand your changes to RexsterTitanServer?

Sure. The diff makes it hard to see what's happening.  I found that placing `ReporterConfig` setup first (before the server objects were created) made things work for me. However, `ReporterConfig` needs a `RexsterApplication` parameter, which in turn needs the graph object. The graph object is created in the `start()` method.

So, to get everything into the correct order, I had to move `RexProRexsterServer` and `HttpRexsterServer` creation out of the constructor, and into the `start()` method.  It's completely possible there are better ways to do this, but I just stopped once I had it working.

dalaro:
Thanks.  I looks OK to me, though I don't have a deep understanding of what's going on here.  Do you think calling `reporterConfig.enable()` before creating server objects is the key point?

I cherry-picked just your changes on RexsterTitanServer and I'm running tests locally now.  Should be pushed shortly...

dalaro:
Pushed RexsterTitanServer changes in f09aa4d82745a7831ba84afcafc7dc53d6671bb3 and slightly different and more restrictive Eclipse .gitignore changes in b68ff656a16d69f315eea8bc3d7cbcf960598e74.  What Eclipse version/feature produces *.prefs?

zachkinstner:
> Do you think calling `reporterConfig.enable()` before creating server objects is the key point?

Yes, I think so.  Practically speaking, I'm not sure that it matters -- the various parameter and logical dependencies surrounding the `reporterConfig.enable()` call are forcing the order of setup/creation.

> What Eclipse version/feature produces *.prefs?

Not sure, entirely... I've only been working with Java for the past few weeks.  I'm using "Eclipse IDE for Java Developers, Version: Juno Service Release 2" with the Maven plugin installed.

spmallette:
Given the version at which you guys are working, calling `enable()` is important as the reporters are not started in the individual `RexsterServer` implementations.  I'd expect some more change around this feature given some work i'm currently doing in a branch...you likely shouldn't have to call `enable()` anymore once we have a stable release of 2.4.0.

dalaro:
Thanks, @spmallette.

I think the contents of this request have been completely merged into titan03, so I'm closing it.  Please do open another request if I've missed something.

PR #287, opened 2013-05-22
Allow IN-unique and OUT-unique edges to have primary keys

zachkinstner:
I tested this change on a local Berkeley database and on my EmbeddedCassandra cluster -- it was successful for both. I'm sure this would need more thorough review, but these early tests show that it works for simple scenarios.

The following use case explains why this change is important to me:

> One million Factors all "use" Artifact X. Each Factor may "use" only one Artifact (OUT unique). Along a user's traversal path, they hit Artifact X. They want to find all Factors which "use" Artifact X and have a FactorType property value of 123.  If I added FactorType as a primary key to the "use" edge, the documentation shows that I would see a dramatic performance improvement. However, because the "use" edge is OUT unique, I can't add a primary key.

The test results show that this _does_ provide a dramatic performance improvement.

I can't get Titan tests to pass in Eclipse (for any branch), so I don't know if this change will cause any existing tests to fail.

Tests and details:
https://gist.github.com/zachkinstner/5630877

Discussion on this topic:
https://groups.google.com/forum/#!topic/aureliusgraphs/KFVtfC9Xy94

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/types/StandardTypeMaker.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/types/StandardTypeMaker.java
@@ -78,7 +78,7 @@ public class StandardTypeMaker implements TypeMaker {
         checkSignature(signature);
         Preconditions.checkArgument(Sets.intersection(Sets.newHashSet(primaryKey),Sets.newHashSet(signature)).isEmpty(),
                 "Signature and primary key must be disjoined");
-        if ((isUnique[0] || isUnique[1]) && !primaryKey.isEmpty()) throw new IllegalArgumentException("Cannot define a primary key on a unique type");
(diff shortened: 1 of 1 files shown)


zachkinstner:
From the discussion thread:

> I think I've read that edges are stored twice (one copy with each vertex), so perhaps one end's "unique" optimizations don't affect the other end. [...] I'm optimistic about this change, but it still feels like there's some big, bad technical reason why this can't be done.

zachkinstner:
This particular line of code was added here: https://github.com/thinkaurelius/titan/commit/5e4ebed47a44f33ee6b90e35ab380934e849499f#L5R79

mbroecheler:
Hey Zack, sorry I had to let this sit for a while because we were refactoring the disk representation for Titan 0.4.0. This addresses an important (for speed purposes) bug and I will make sure its being addressed in the next release. Until then, I will leave this PR open.

Issue #298, opened 2013-06-05
RexsterTitanServer ignores <script-engines> in  rexster.xml


vantonio:
`EngineController.configure(-1, this.rexsterConfig.getString("script-engine-init", null));`
in file RexsterTitanServer.java,  ignores the <script-engines> tag.

spmallette:
@dalaro this is a good fix as we have never allowed the configuration of multiple script engines in Titan Server before.    i assume we would want to just merge this to master.  any objections?

btw, @vantonio thanks for the pull request.

dalaro:
This caused an NPE for null script-engines values.  At least one of the unit tests based on `RexsterServerClientTest` died immediately on this NPE.  I modified the code slightly to workaround the NPE by falling back on the old behavior when script-engines is unset and pushed 110476b02d9b39451fff242a7fbb0aeae2efc082 to titan03.

Do you think this needs to go into master immediately?  It seems at least as much a feature as a bug.  Maybe I just don't understand.

spmallette:
It's not really a bug.  it's more of a feature omission, which was somewhat intentional when we first got started. i'll make a couple of adjustments to what you have in titan03.  besides, i busted some stuff when i deployed latest rexster to sonatype a few hours ago...

spmallette:
Committed this fix... 0f88d61d90d95ef1dcfd82aec48cf4be6738813b ...should now run with rexster 2.4.0-SNAPSHOT.

vantonio:
Thank you guys:-)

dalaro:
Thank you.  I think this and the related PR #299 are resolved and I'm closing them.  Please reopen if I missed something.

PR #299, opened 2013-06-05
fixed Issue #298

vantonio:
reads the <gremlin-engines> tag, and pass it to EngineController.


dalaro:
Pulled related fix into titan03 as part of #298

PR #323, opened 2013-07-16
Updates for Latest Blueprints Interfaces

zachkinstner:
For my Fabric/RexConnect projects, I wanted to try/test the latest RexPro changes with my existing Titan setup. However, Titan wouldn't compile with the latest Blueprints/Rexster. Starting from the `standard-index-experimentation` branch (seemed like the best option), I implemented the changes shown here.

I'm not sure if these changes will be useful/helpful for Titan dev, but I thought I'd share them. Perhaps it would make a decent starting point for the upgrade/refactor to the latest Blueprints. Titan will compile with these changes, although I had to omit `titan-tests` from `pom.xml` to accomplish this -- it has a variety of build failures that would require more extensive code changes to resolve.

This build of Titan executes fine for me locally, and RexPro/RexConnect are working as expected. It also passes my entire suite of Fabric integration tests, which include a variety of index scenarios (both "standard" and with ElasticSearch).

### DIFF
--- a/doc/javascript/editor/langs/creole.js
+++ b/doc/javascript/editor/langs/creole.js
@@ -22,7 +22,7 @@ var Creole = {
                             },
                             
   'function-hr'     :       {
-                              append: "\n\n----\n\n";
+                              append: "\n\n----\n\n"
                             },
   
   'function-ul'     :       { 
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     </properties>
     <modules>
         <module>titan-core</module>
-        <module>titan-test</module>
+        <!--module>titan-test</module-->
         <module>titan-berkeleyje</module>
         <module>titan-cassandra</module>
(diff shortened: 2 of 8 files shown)


dalaro:
Thanks!

I think that Matthias has also been considering dramatically refactoring this code, which is part of the reason that I idled my standard-index-experimentation branch.  Let's wait for his input before proceeding.

mbroecheler:
@zachkinstner: thanks for jumping in on this. Check out the blueprints-2.4.0 branch. I think it should contain what you want. There is some more work to come on that branch (query optimization mostly) but as it currently stands (commit 64a3847562d02b198c868bb01a14572655b1b827) it is compatible with the most recent blueprints.

zachkinstner:
I updated to the `blueprints-2.4.0` branch. So far, it's working fine, and all of my Fabric integration tests pass.

PR #362, opened 2013-09-05
Elasticsearch/Lucene should index fields by their actual name

erickt:
This pull request makes a backwards incompatible change to make Titan index fields by their proper `String` field name. Before this pull request, Titan instead indexed fields by an internal unique `Long` identifier, which makes it difficult to directly access Elasticsearch to perform queries that are not general purpose enough to be exposed through the Titan API.

An alternative to this approach would be to directly expose discovering the `Long` identifier for the fields, and then an end user could handle converting a string field name to and from that identifier, but I feel that would be pretty painful for the user. Another way would be to expose a "raw query" API, but I feel it would be difficult to make a generic interface. It would be ultimately easier to just to get a reference to the underlying index and query directly against that. I plan to implement this in a later pull request.

There are two main downsides to this approach. First, is that this will force users of Titan 0.3.x to reindex their graphs. Second, there _may_ be some overhead by storing the full string instead of an encoded integer, but lucene is pretty efficient about storing documents, so I don't think it will add a significant overhead to Titan's indexing support.

For more details, you can refer to this [gist](https://gist.github.com/erickt/6444959) to show the benefits of this change.

Finally, this pull request does also include some minor code cleanup.

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/diskstorage/indexing/IndexTransaction.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/diskstorage/indexing/IndexTransaction.java
@@ -32,26 +32,26 @@ public class IndexTransaction implements TransactionHandle {
         this.mutations = null;
     }
 
-    public void add(String store, String docid, String key, Object value, boolean isNew) {
-        getIndexMutation(store,docid,isNew,false).addition(new IndexEntry(key,value));
+    public void add(String storeName, String docId, String key, Object value, boolean isNew) {
+        getIndexMutation(storeName,docId,isNew,false).addition(new IndexEntry(key,value));
     }
 
-    public void delete(String store, String docid, String key, boolean deleteAll) {
-        getIndexMutation(store,docid,false,deleteAll).deletion(key);
+    public void delete(String storeName, String docId, String key, boolean deleteAll) {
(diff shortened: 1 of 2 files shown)


mbroecheler:
The reason we used the long ids to begin with is that it makes it easier to manage the schema across the different data stores. While not yet supported, we eventually want to allow users to change type definitions, including changing names of property keys. According to what I read about Lucene/ES, that would be an extremely expensive operation.

Rather, I wanna go the route of exposing a string query API where users can pass arbitrary (string) based queries through to the index and Titan translates fields to ids.

mbroecheler:
Using an encoded type id has the advantage that:
1) Renaming the type does not impact the indexing backend
2) Naming restrictions of the indexing backend does not impact Titan

The downside is that directly accessing the indexing backend is cumbersome. Hence, we plan to provider helper methods to make this easier in the future which is why this ticket remains open until this has been implemented.

mbroecheler:
@erickt We extended Titan's Lucene/ElasticSearch capabilities by parsing Lucene style queries and passing them into Lucene. That way, a user need not be concerned with how the names show up in the index and we can continue using the ids which gives us more flexibility and better upgradability in the future.

Please have a look at this and tell us if this may work for you.

mbroecheler:
fixed in titan05

Issue #312, opened 2013-06-26
Cassandra backed support through native transport and cql


devsprint:
I'm looking to implement storage support for Cassandra, using CQL and native transport protocol. 

I do have a draft version that is working  but there are still improvements to add.

I have implemented as a separate module, to not mess with the dependencies.

If is it something that it could be interesting to Titan community, then, I will make a pull request.

Let me know your thoughts.

mbroecheler:
Hey Gabriel,

what is the performance of using cql vs thrift? What would be the benefit
of going over cql?

Cheers,
Matthias

On Wed, Jun 26, 2013 at 2:06 PM, Gabriel Ciuloaica <notifications@github.com

> wrote:
> 
> I'm looking to implement storage support for Cassandra, using CQL and
> native transport protocol.
> 
> I do have a draft version that is working but there are still improvements
> to add.
> 
> I have implemented as a separate module, to not mess with the dependencies.
> 
> If is it something that it could be interesting to Titan community, then,
> I will make a pull request.
> 
> Let me know your thoughts.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/issues/312
> .

## 

Matthias Broecheler
http://www.matthiasb.com

devsprint:
I think that the main benefit will be better performance by using the native protocol and cql instead of thrift. I wil come back in a week with some performance numbers. Is there any specific test/scenario that you'd like to see the numbers for?

Br,
Gabi

mbroecheler:
That would be fantastic? No specific scenario, just some standard vertex
and index queries over a reasonable size dataset.

Thanks,
Matthias

On Thu, Jun 27, 2013 at 9:33 AM, Gabriel Ciuloaica <notifications@github.com

> wrote:
> 
> I think that the main benefit will be better performance by using the
> native protocol and cql instead of thrift. I wil come back in a week with
> some performance numbers. Is there any specific test/scenario that you'd
> like to see the numbers?
> 
> Br,
> Gabi
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/issues/312#issuecomment-20136127
> .

## 

Matthias Broecheler
http://www.matthiasb.com

mbroecheler:
@devsprint Any news on this?

devsprint:
Hi, I just merged with latest titan code. I still have some functional tests failing but I should have something soon.

poelzi:
[...]

PR #364, opened 2013-09-05
Titan with Cassandra CQL

devsprint:
This is the current implementation of CQL support (#312). There are failing unit tests, mostly caused by latest changes in Titan core, that I did not yet implemented.

Some refactoring is still required.

I will continue to push my work to this pull request. 

Any comments/suggestions are welcome. 

I have tried to use same data model, replacing wide rows with cql equivalent, that is table with compound primary key.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,7 @@
         <module>titan-test</module>
         <module>titan-berkeleyje</module>
         <module>titan-cassandra</module>
+        <module>titan-cassandra-cql</module>
         <module>titan-persistit</module>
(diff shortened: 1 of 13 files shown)


mbroecheler:
We have been keeping an eye on Cassandra's support for thrift and it is clear that things are moving in the direction of CQL. That said, there are some features that make it appealing to use thrift for now (less overhead, multi_slice_get).
We will revisit the usage of CQL in the future and keep this PR open until then for inspiration.

tspecht:
Any news on CQL support?

PR #372, opened 2013-09-15
Add Solr as New Index Provider for Titan

jholmberg-cni:
We've been working to integrate Titan and Faunus into our project as its data model fits closely with our requirements. However, we have an existing need to use Solr instead of Lucene or ElasticSearch.  

This fork contains a new Apache Solr based Index Provider that works with the traditional HTTP Solr Server and the SolrCloud Server that runs with Apache Zookeeper. It's still pretty new but it currently passes all tests in the IndexProvider unit tests.  I have another github site containing instructions on how to get Solr set up in your development environment to test out both modes of Solr (HTTP and Cloud).

You can check the instructions out at: https://github.com/tenaciousjzh/titan-solr-cloud-test

If anyone chooses to run Solr as their Index Provider, they'll need to know in advance that Solr does not work quite like Elastic.  They need to configure Solr with the correct schema first before telling Titan about it. The example github project previously mentioned ships with 4 schemas that are used to create Solr cores for each of the different stores mentioned in the unit tests. 

Please feel free to contact me should you be interested in merging this functionality into Titan.  Thanks in advance for your consideration and I look forward to hearing from you.

Best Regards,
Jared Holmberg


jascks:
Are there any plans to merge this pull request?  I'm working in a Solr shop (DataStax Enterprise, DSE) as well, and I would very much like to make use of Solr rather than setup ES etc.  It seems Titan with DSE is an ideal combination with enterprise grade Cassandra and Solr at hand.

Thanks!

jholmberg-cni:
I haven't received any word on plans to merge this pull request yet; but I know that the Titan team is working pretty hard maturing the platform. The folks that I have corresponded with in the past have been very good to work with. 

I have been keeping the pull request up-to-date until about a month ago. Titan is evolving pretty rapidly so I've had to change the code quite a bit to keep up. Just as a heads up, I have only implemented support for SolrHttp and SolrCloud, not Solr Embedded.

I'm currently in the process of building out a replacement to our existing database solution that involves Titan as our primary data store. However, our software has some stringent security requirements so I'm working on locking down access based on a security model similar to Accumulo.  So we don't have this running in production yet and I believe besides the unit tests, the Solr provider would need some good integration tests before I felt comfortable with it.  If you check out the titan-solr-cloud-test project mentioned when I created the pull request, I think that could possibly be used as a basis for getting good integration tests set up that could run on SolrCloud.

I'd be happy to collaborate with you on this where time permits.

jascks:
Hi Jared, thanks for responding.  I’m glad to hear the Titan folks are receptive to this, but just busy. :)

Titan is a ways from a 1.0 release from what I can tell. :) I can imagine it takes some effort to keep up with it.  Hopefully the search backend interfaces don’t change too often.

My only concern is SolrHttp as I use the Java SolrJ client library to interact with Solr today, and do not use Solr cloud. DataStax Enterprise is useful to provide the replication, redundancy, no single point of failure etc.

I’m hoping I’ll be taking a more serious look at all this during the next few two-week sprints I have coming up.  This is a longer term investment than that of course, and as much as I see the potential of Titan with our new data model, the Solr search backend has been sticking out like a sore thumb.  It would be great to get that resolved as soon as possible.

I’m happy to collaborate with you as well!  I’ll start working with titan-solr-cloud-test and see how it goes.  Perhaps I’ll be submitting pull request to you. :)

Cheers,

Jeff

jholmberg-cni:
Pull requests are definitely welcome! I definitely would like to get your
perspective and any improvements you'd like to make to the code.
Cassandra/DataStax is a fantastic solution, and I think its the most mature
engine that Titan is running on. We have a need to use HBase or Accumulo
but if I had my druthers, I'd go for Cassandra/DataStax.  It's got better
performance, and I like that it has no single point of failure.

Look forward to your participation!
-Jared

jascks:
Thanks Jared. It should be good!

mbroecheler:
Hi Jared et al,
thanks for your patience regarding this pull request. A Solr adapter for Titan is definitely on the roadmap but we are becoming more wary regarding large pull requests because it often boils down to us having to maintain them ;-)
For this purpose, we started a "Titan experimental" repository where we can play with things. 
So, we will keep an eye on this and test it internally and hopefully merge it soon.

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
```diff
@@ -0,0 +1,688 @@
...
+    @Override
+    public Iterable<RawQuery.Result<String>> query(RawQuery query, KeyInformation.IndexRetriever informations, TransactionHandle tx) throws StorageException {
+        List<RawQuery.Result<String>> result = new ArrayList<RawQuery.Result<String>>();
+        String core = query.getStore();
+        String keyIdField = keyFieldIds.get(core);
+        SolrServer solr = this.solrServers.get(core);
+        SolrQuery solrQuery = new SolrQuery();
+        solrQuery.setQuery("*:*");
+        solrQuery.addFilterQuery(query.getQuery());
+        solrQuery.addField("*");
```
jascks: Why retrieve all fields when only the score and the key ID field are used below?  Seems this would cause unnecessary overhead particularly for stores with lots and/or multi-valued fields.

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
```diff
@@ -0,0 +1,688 @@
...
+                    //Handle any deletions
+                    if (mutation.hasDeletions()) {
+                        if (mutation.isDeleted()) {
+                            log.trace("Deleting entire document from Solr {}", docId);
+                            deleteIds.add(docId);
+                        } else {
+                            deleteIndividualFieldsFromIndex(solr, keyIdField, docId, mutation);
+                        }
+                    } else {
+                        Set<String> fieldDeletions = Sets.newHashSet(mutation.getDeletions());
```
jascks, line 238: Perhaps a dumb question as I've never used Guava, but from what I can tell fieldDeletions is a Set in order to remove potential duplicates? Does invoking its remove() method on line 241 also affect the underlying List of deletions referenced by the Mutation instance?  If not, then fieldDeletions is not otherwise used, and then goes out of a scope with the else block. Did anything useful happen?

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
```diff
@@ -0,0 +1,688 @@
...
+                    commitDocumentChanges(solr, updateDocuments, isLastBatch);
+                }
+            }
+        } catch (Exception e) {
+            throw storageException(e);
+        }
+    }
+
+    private void deleteIndividualFieldsFromIndex(SolrServer solr, String keyIdField, String docId, IndexMutation mutation) throws SolrServerException, IOException {
+        Set<String> fieldDeletions = Sets.newHashSet(mutation.getDeletions());
```
jascks, line 295: Unlike my comment above, I believe this Set is created in order to remove potential duplicates, since fieldDeletions is used in the for loop below.

titan-core/src/main/java/com/thinkaurelius/titan/diskstorage/Backend.java
```diff
@@ -477,6 +477,7 @@ public void clearStorage() throws StorageException {
         put("lucene", "com.thinkaurelius.titan.diskstorage.lucene.LuceneIndex");
         put("elasticsearch", "com.thinkaurelius.titan.diskstorage.es.ElasticSearchIndex");
         put("es", "com.thinkaurelius.titan.diskstorage.es.ElasticSearchIndex");
+        put("solr", "com.thinkaurelius.titan.diskstorage.solr.SolrIndex");
```
jascks, line 480: This makes it difficult for search backends to be independent from Titan itself.  That is, suppose I subclass the new SolrIndex class as say, DseSolrIndex to make some accommodations for DataStax Enterprise.  Then I'd have to add another map entry in the Titan project itself so this IndexProvider implementation would be recognized.

I admit, I'm just now looking at Backend.java in more detail, so maybe I'm missing something. Would it make sense to define annotations such as \@IndexProvider and \@StorageManager and then Titan can scan for those and register the components actually present on the classpath?  Or I suppose implementations of the IndexProvider could be located, and then invoke a new registerProvider() method which returns the 'shortcut(s)' by which the provider can be referenced?

mbroecheler, line 480: Hi Jeff,

now, you don't need to add these to Backend.java. These are just shortcuts
for commonly used IndexProviders. You can always provide the full class
name in the configuration without making any changes to Titan core.
So, the code you are referencing is just syntactic sugar.

On Mon, Jan 20, 2014 at 6:05 PM, Jeff Schmidt notifications@github.comwrote:

> In
> titan-core/src/main/java/com/thinkaurelius/titan/diskstorage/Backend.java:
> 
> > @@ -477,6 +477,7 @@ public void clearStorage() throws StorageException {
> >          put("lucene", "com.thinkaurelius.titan.diskstorage.lucene.LuceneIndex");
> >          put("elasticsearch", "com.thinkaurelius.titan.diskstorage.es.ElasticSearchIndex");
> >          put("es", "com.thinkaurelius.titan.diskstorage.es.ElasticSearchIndex");
> > -        put("solr", "com.thinkaurelius.titan.diskstorage.solr.SolrIndex");
> 
> This makes it difficult for search backends to be independent from Titan
> itself. That is, suppose I subclass the new SolrIndex class as say,
> DseSolrIndex to make some accommodations for DataStax Enterprise. Then I'd
> have to add another map entry in the Titan project itself so this
> IndexProvider implementation would be recognized.
> 
> I admit, I'm just now looking at Backend.java in more detail, so maybe I'm
> missing something. Would it make sense to define annotations such as
> \@IndexProvider and \@StorageManager and then Titan can scan for those and
> register the components actually present on the classpath? Or I suppose
> implementations of the IndexProvider could be located, and then invoke a
> new registerProvider() method which returns the 'shortcut(s)' by which the
> provider can be referenced?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/372/files#r9025422
> .

## 

Matthias Broecheler
http://www.matthiasb.com

jascks, line 480: Thanks Matthias, that is good to know.  From what I can tell from the properties being loaded in the static initializer, I can define:

index.solr=com.thinkaurelius.titan.diskstorage.solr.SolrIndex

That adds the shorthand "solr" to the registered index providers.  Then with that class and its dependencies on the classpath, I should be good to go? I can then define:

storage.index.search.backend=solr

And then when I do something like:

graph.makeKey('symbol').dataType(String.class).indexed('search', Vertex.class).make()

I'll then be working with Solr?

Thanks,

Jeff

mbroecheler, line 480: No, you cannot register additional shorthands that way. Or rather, you can,
but we will remove that since its causing dependency issues due to
classpath ordering issues.

Best you do:

storage.index.search.backend = com.thinkaurelius.titan.diskstorage.solr.
SolrIndex

That always works.

On Fri, Jan 24, 2014 at 4:21 AM, Jeff Schmidt notifications@github.comwrote:

> In
> titan-core/src/main/java/com/thinkaurelius/titan/diskstorage/Backend.java:
> 
> > @@ -477,6 +477,7 @@ public void clearStorage() throws StorageException {
> >          put("lucene", "com.thinkaurelius.titan.diskstorage.lucene.LuceneIndex");
> >          put("elasticsearch", "com.thinkaurelius.titan.diskstorage.es.ElasticSearchIndex");
> >          put("es", "com.thinkaurelius.titan.diskstorage.es.ElasticSearchIndex");
> > -        put("solr", "com.thinkaurelius.titan.diskstorage.solr.SolrIndex");
> 
> Thanks Matthias, that is good to know. From what I can tell from the
> properties being loaded in the static initializer, I can define:
> 
> index.solr=com.thinkaurelius.titan.diskstorage.solr.SolrIndex
> 
> That adds the shorthand "solr" to the registered index providers. Then
> with that class and its dependencies on the classpath, I should be good to
> go? I can then define:
> 
> storage.index.search.backend=solr
> 
> And then when I do something like:
> 
> graph.makeKey('symbol').dataType(String.class).indexed('search',
> Vertex.class).make()
> 
> I'll then be working with Solr?
> 
> Thanks,
> 
> Jeff
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/372/files#r9146715
> .

## 

Matthias Broecheler
http://www.matthiasb.com

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
jholmberg-cni: Hi Jeff,

Good catch. As I was putting this together, I was learning Solr at the same
time. Having another pair of eyes on this helps a lot.
I have changed the code from :

solrQuery.addField("*");
solrQuery.addField("score");

to:

solrQuery.addField(keyIdField);
solrQuery.addField("score");

It's been updated and pushed to github.

-Jared

titan-core/src/main/java/com/thinkaurelius/titan/diskstorage/Backend.java
jascks, line 480: Good to know. Thanks Matthias!

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
jascks: Sounds good!  Thanks Jared!

## Jeff

Jeff Schmidt
535 Consulting
jas@535consulting.com
http://www.535consulting.com

erickt:
@tenaciousjzh: I've been working on this patch here: https://github.com/lab41/titan/tree/solr. It's not quite done yet, but so far I've managed to get embedded mode mode to work well, fixed up geoshape handling, and fleshed out some tests and error handling. Next up is getting full text searches to work like elasticsearch, and explore supporting solr schema-less indexing.

jascks:
@tenaciousjzh, that is great news!  Thank you!

Jeff

jholmberg-cni:
Awesome Erick!  That's great! That's going to make testing go a lot better
with embedded solr working. I'm admittedly less experienced with
schema-less indexing so best of luck in figuring that out. In talking with
some Solr experts I know, they tended to prefer using schemas. That said,
having that capability as an option would definitely fit better with Titan.
-Jared

fancyerii:
will it possible to be a standard feature of titan rather than a test branch? compiling from source and keep up with latest version for normal user is difficult.

jholmberg-cni:
Unfortunately, that is not really under my control. As Matthias mentioned,
pulling in a new provider is a fairly big undertaking. I know the Aurelius
team is working on many features for their upcoming release(s). It could be
worthwhile to check in with them and see where this falls on their roadmap.

For now, compiling your own may be your best option. It may be useful for
you to deploy the compiled project to your own local maven repository. That
lets you add the maven dependency to your project and saves you from having
to build it every time.

Cheers!
-Jared

mbroecheler:
We will definitely make the SOLR adapter part of the TItan codebase, but unfortunately I cannot yet promise when that will happen. We have our hands full atm with Titan 0.5 and after that we will adjust our roadmap and see where the SOLR adapter lands. I will keep you posted.

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
```diff
@@ -0,0 +1,689 @@
...
+
+                    //Handle any deletions
+                    if (mutation.hasDeletions()) {
+                        if (mutation.isDeleted()) {
+                            log.trace("Deleting entire document from Solr {}", docId);
+                            deleteIds.add(docId);
+                        } else {
+                            deleteIndividualFieldsFromIndex(solr, keyIdField, docId, mutation);
+                        }
+                    } else {
```
xedin, line 237: Can you elaborate on why is this "else" block even required? Next if condition is handling hasAdditions already...

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
```diff
@@ -0,0 +1,689 @@
...
+                sb.append(fieldToDelete + ",");
+            }
+            log.trace("Deleting individual fields [{}] for document {}", sb.toString(), docId);
+            solr.add(doc);
+            solr.commit();
+
+        }
+    }
+
+    private void commitDocumentChanges(SolrServer server, Collection<SolrInputDocument> documents, boolean isLastBatch) throws SolrServerException, IOException {
```
xedin, line 313: I think this should be named maybeCommitChanges() because it only does commit when BATCH_SIZE or reached or it was a last batch.

titan-solr/src/main/java/com/thinkaurelius/titan/diskstorage/solr/SolrIndex.java
```diff
@@ -0,0 +1,689 @@
...
+            for (SolrInputDocument d : documents) {
+                Collection<String> fieldNames = d.getFieldNames();
+                for (String name : fieldNames) {
+                    log.error(name + ":" + d.getFieldValue(name).toString());
+                }
+            }
+        }
+    }
+
+    private void commitDeletes(SolrServer server, List<String> deleteIds, boolean isLastBatch) throws SolrServerException, IOException {
```
xedin, line 346: The same (as in "commitDocumentChanges") name name change applies here too

xedin:
@mbroecheler @tenaciousjzh Overall this looks solid (needs some dad code cleanup and renaming). But the concern I want to raise here is related to the use of Solr "Atomic Updates", which in practice doesn't really scale because of the single file commit log design that has to be read for every operation (update or get), which makes it a bottleneck. So I would say we either support two modes - read/write on the whole documents (which requires read for ever update we are going to do) and "Atomic Updates" for the small use-cases or leave this as-is but mark as experimental. WDYT?

mbroecheler:
@xedin thanks for taking a look. Is it possible to make this configurable? And can you elaborate on what a more scalable write model would look like?

erickt:
@xedin: how's the patch coming along? Is there anything I can do to help?

mbroecheler:
@erickt Pavel managed to get a version working that uses schemaless mode. This is somewhat less efficient than actively managing the schema but seems to be the only reliable way to make it work with SolrCloud.

dalaro:
I think the changes in this PR have all made it into titan05.  I also think schemaless mode was recently implemented via dynamic field mapping support after the merge into titan05, and it looks like atomic update work has moved onto #837.

I don't see any remaining "next steps" in this PR, so I'm closing it.  But please reopen if I've missed something and closed it prematurely.

PR #377, opened 2013-09-23
Accumulo Storage Backend

deprit:
This fork contains a new Accumulo v1.4.3 storage backend for Titan.

The Accumulo backend is organized as two modules: core and iterators. Titan-accumulo-core contains classes implementing the Titan storage API. Titan-accumulo-iterators provides a column filter used for slice queries, and the Accumulo storage manager may be configured to run this iterator on client or server side.

This initial version does not make use of Accumulo's cell visibility mechanism.


mbroecheler:
Leaving this open for consideration / inspiration but we won't be able to merge in an entire storage backend without continued support by the developer.
We encourage that future storage backend contributions are held in the developer's repository before we consider merging it into the titan repo.

PR #397, opened 2013-09-30
add configuration options to control Cassandra compression

xedin:


mamccr:
Hey, sorry again for the overlap here.  Pull Request 405 is basically just like this, but allows you to specify an alternative compression algorithm for Cassandra instead of just turning off compression all together.  I've tested that this works on the IBM mainframe using the jre provided there (this was what drove me to independently look at the same issue).  Please take a look and see if that is a viable alternative to this pull request.  The other option is to sort of merge the two of these and have both a "skip compression all together" option and, if that one isn't set, fall back to the new option in my pull request.  I can put that together if that's what you would prefer.

xedin:
Well, we don't have anything better than Snappy in Cassandra and most certainly wouldn't have it in near future, so I'm not sure if making it configurable is useful on Titan side especially as it's not the only way to set things up.

mamccr:
Hey, I agree that there isn't anything better than Snappy in Cassandra.  But I think some of the other alternatives are better than no compression at all for platforms that aren't supported by Snappy.  There's also a true java Snappy implementation floating around out there (https://github.com/dain/snappy) that's pretty close in performance which could run on all those platforms.  My proposed change would let folks use any of those if they wanted.  Does that make sense?  Also, without the change, I don't know of any way to set up compression for sstable compression when setting up Cassandra through Titan, since it's currently hardcoded to SnappyCompressor in Titan.  Does that make sense?  Is there something else I'm missing or overlooking?

xedin:
@mamccr as compression is done on CF level you can change it anytime you want using Cassandra's build-in tools such as CLI by updating CF with compression_options you want.

dalaro:
Thanks for the code.  I merged this and #405.  I made some additional changes, like javadoc for #405's new config key, bringing the config key strings into agreement on a naming pattern, and rewriting CompressionParameters handling in the embedded adapter, which was somewhat out-of-date and incomplete.

The defaults remain unchanged: compression enabled, Snappy, 64 kb chunk length.  But now those can all be overridden through Titan config options when any of the Cassandra backends is configured:
- storage.compression.enabled
- storage.compression.sstable_compression
- storage.compression.chunk_length_kb

(assuming storage.backend = a cassandra implementation)

I also made tests to verify that these options actually work on each of the Cassandra backends, but I got sidetracked with the getKeys method in the Cassandra adapters.  I haven't cleaned them up for a push yet.

dalaro:
Added compression configuration tests as part of a649b7a3d40eae5fb7cbbe6b37988a002b6a1ed2 in [AbstractCassandraKCVSTest](https://github.com/thinkaurelius/titan/blob/a649b7a3d40eae5fb7cbbe6b37988a002b6a1ed2/titan-cassandra/src/test/java/com/thinkaurelius/titan/diskstorage/cassandra/AbstractCassandraKeyColumnValueStoreTest.java)

PR #408, opened 2013-10-01
Add configuration option for the max number of results returned by an ElasticSearch index query

akeffalas:
There may be times when limiting the result set size of an ElasticSearch index query to the currently fixed value of 100,000 results is not desired.  If you have a large dataset hosted on a machine capable of handling a large number of results the ElasticSearchIndex will currently 'truncate' your result set to a fixed value that may be too small (or large).  This pull request allows for the max result set size to be configured.

This change was tested with the following configuration:

```
Configuration conf = new BaseConfiguration();
conf.setProperty("storage.backend","berkeleyje");
conf.setProperty("storage.directory","/tmp/titan_test");

conf.setProperty("storage.index.search.backend", "elasticsearch");
conf.setProperty("storage.index.search.hostname", "localhost");
conf.setProperty("storage.index.search.client-only", "true");
conf.setProperty("storage.index.search.max-result-set-size", "500000");

TitanGraph graph = TitanFactory.open(conf);
```


mbroecheler:
Merged in https://github.com/thinkaurelius/titan/commit/cfecb0c85cf2539eb2bc01c90a83e7f3572d57d6

PR #441, opened 2013-10-19
Add support for search within polygon Geoshape

kainosnoema:
The current version of ElasticSearch supports the GeoPolygonFilter, so it was trivial to add support for polygons with an arbitray number of boundary points. Works great. I added a test and the suite passes.

The trickiest part of this is the interface to easily create a polygon Geoshape with both Java and Groovy syntax. This is where I ended up (seemed consistent with the other constructors), but I'm open to suggestions:

``` groovy
g.V.has('location', WITHIN, Geoshape.polygon(37.0,22.0, 40.0,22.0, 40.0,24.0, 37.0,24.0))
```

We could support `ArrayList<double>(2)` arguments to make the Groovy a bit nicer, but it would increase the complexity. Thoughts?

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/core/attribute/Geoshape.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/core/attribute/Geoshape.java
@@ -257,6 +257,39 @@ public class Geoshape {
     }
 
     /**
+     * Constructs a new polygon shape which is identified by its coordinates
+     * @param coordinates
+     * @return
+     */
+    public static final Geoshape polygon(final float ... coordinates) {
+        Preconditions.checkArgument(coordinates.length % 2 == 0, "Odd number of coordinates provided");
+        Preconditions.checkArgument(coordinates.length >= 6, "Too few coordinate pairs provided");
+        float[] latitudes = new float[coordinates.length / 2];
+        float[] longitudes = new float[coordinates.length / 2];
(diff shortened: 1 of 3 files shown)


kainosnoema:
I'm not sure exactly why the Travis build is failing, but it doesn't seem to be related to this commit: `[INFO] Titan-Lucene: Indexing Support .................... FAILURE [14.386s]`. The master branch branch was passing for me locally just fine prior to pushing this. Is this a known issue at the moment?

mbroecheler:
Yes, travis is overwhelmed by the complexity of the Titan build and will
fail due to time-outs. We have tried to navigate this by splitting the
build into a small and large build so it can complete on time, but Travis
is very restrictive and the expansion of Titan has caused our test coverage
and therefore unit test time to increase.

On Fri, Oct 18, 2013 at 9:57 PM, Evan Owen notifications@github.com wrote:

> I'm not sure exactly why the Travis build is failing, but it doesn't seem
> to be related to this commit: [INFO] Titan-Lucene: Indexing Support
> .................... FAILURE [14.386s]. The master branch branch was
> passing for me locally just fine prior to pushing this. Is this a known
> issue at the moment?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/441#issuecomment-26643518
> .

## 

Matthias Broecheler
http://www.matthiasb.com

Issue #439, opened 2013-10-18
multiQuery() fails to process has clause.


haebin:
I don't know how to make has() work.
I tried several combinations but it just doesn't work.

rexster[groovy]> g.multiQuery().addAllVertices(r).direction(OUT).labels('like').has('lid').vertices()
==>An error occurred while processing the script for language [groovy]. All transactions across all graphs in the session have been concluded with failure: java.util.concurrent.ExecutionException: javax.script.ScriptException: javax.script.ScriptException: java.lang.IllegalArgumentException

rexster[groovy]> g.multiQuery().addAllVertices(r).direction(OUT).labels('like').has('lid','1').vertices()
==>An error occurred while processing the script for language [groovy]. All transactions across all graphs in the session have been concluded with failure: java.util.concurrent.ExecutionException: javax.script.ScriptException: javax.script.ScriptException: java.lang.IllegalArgumentException

rexster[groovy]> g.multiQuery().addAllVertices(r).direction(OUT).labels('like').has('lid',Compare.EQUAL,'1').vertices()
==>An error occurred while processing the script for language [groovy]. All transactions across all graphs in the session have been concluded with failure: java.util.concurrent.ExecutionException: javax.script.ScriptException: javax.script.ScriptException: java.lang.IllegalArgumentException

PR #442, opened 2013-10-19
multiQuery has(), g.V('pk', byteArray) bug patches.

haebin:
multiQuery has() throwing IllegalArgumentException bug fix.
https://github.com/thinkaurelius/titan/issues/439 bug fix patch.

Cannot query g.V('primaryKey', byteArray) because byte[] is converted to List<Byte>
https://github.com/thinkaurelius/titan/issues/443 bug fix patch.


mbroecheler:
I had already fixed #439 separately (and differently). Just committed those changes.
The ByteArrayHandler commit shows a great level of understanding of Titan's internals. Good job. Can you pull this commit into a separate pull request? And I guess we need to introduce these handlers for all primitive arrays since this problem will be the same for int[], short[], etc when used through groovy.

haebin:
Thank you, Matthias.
Yet, I just fixed the obvious ones by remote debugging since I needed working version of mutliQuery fast.
But, kinda curious, for https://github.com/thinkaurelius/titan/issues/439, did you have to change all the things?
I thought it was just caused by unhandled query validation. I guess, I will have to read more about your codes.

P.S. Holy cow, separating pull requests is really difficult. Make new branch and cherrypick and commit, push, and pull request. OMG. XD
https://github.com/thinkaurelius/titan/pull/452

PR #444, opened 2013-10-22
Fix memory leaks in LRUVertexCache and StandardTitanTx

jkschneider:
The recently introduced LRUVertexCache slowly consumes tenured heap over time.  Transaction commit and rollback were not completely releasing cache resources used by the transaction during its lifetime.

I believe it is wise to build the vertex cache with softValues() along with a maximum capacity.  This way, applications can recover from misjudgments in setting the transaction cache capacity rather than throwing excessive GC overhead exceptions.

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/transaction/StandardTitanTx.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/transaction/StandardTitanTx.java
@@ -181,7 +181,7 @@ public class StandardTitanTx extends TitanBlueprintsTransaction {
             public int weigh(IndexQuery q, List<Object> r) {
                 return 2 + r.size();
             }
-        }).concurrencyLevel(concurrencyLevel).maximumWeight(config.getIndexCacheWeight()).build();
+        }).concurrencyLevel(concurrencyLevel).maximumWeight(config.getIndexCacheWeight()).softValues().build();
 
         uniqueLocks = UNINITIALIZED_LOCKS;
         deletedRelations = EMPTY_DELETED_RELATIONS;
@@ -966,7 +966,9 @@ public class StandardTitanTx extends TitanBlueprintsTransaction {
     }
 
     private void close() {
-        //TODO: release non crucial data structures to preserve memory?
+        vertexCache.close();
+        indexCache.invalidateAll();
+        indexCache.cleanUp();
         isOpen = false;
     }
 
--- a/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/transaction/vertexcache/LRUVertexCache.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/transaction/vertexcache/LRUVertexCache.java
@@ -1,7 +1,5 @@
 package com.thinkaurelius.titan.graphdb.transaction.vertexcache;
 
-import com.carrotsearch.hppc.ObjectContainer;
-import com.carrotsearch.hppc.cursors.ObjectCursor;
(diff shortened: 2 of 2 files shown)


dalaro:
The only references to LRUVertexCache are private final fields in StandardTitanTx.  Once all references to a transaction go out of scope, so too do references to its LRUVertexCache.  Then the transaction and its cache should be GC'ed.

The issue description seems to suggest closed transactions getting retained somewhere they shouldn't (maybe in Titan core or maybe in the app -- could be anywhere), and LRUVertexCache just shows up as the biggest heap consumer because it's the memory-hungriest reference on each transaction.  But if this is the problem then the patch doesn't completely fix the leak, it just slows it down by freeing the fattest part of the leaking transactions.

This could also be a symptom of an extremely long lived transaction accumulating an ever-larger LRUVertexCache.  I don't think we ever call cleanUp() on the vertex cache, and Guava doesn't do it automatically by default.

What do you think?  Am I missing something?

dalaro:
Correction: the "ever-larger LRUVertexCache" scenario that I suggested appears impossible, since each LRUVertexCache is created with a maximumSize argument that defaults to 20000.

jkschneider:
I know it's hard to believe at first, but I have found that closing the transaction does not allow the cache contents to be GC'ed.  I could try to go back and write a test to prove this out.  

One of my instances was consistently throwing excessive GC overhead exceptions under heavy load until this fix was applied.  This instance had quite a bit of RAM allocated to it, and you could watch the tenured heap gradually growing until full GCs happened multiple times per query.

These transactions were all very short lived (e.g. performed one query operation and then commit was called).

MAT revealed the following:

image

mbroecheler:
Could it be possible that you are holding on to vertex or edges object that
were created in an older transaction? Even if you close such a transaction,
the transaction object will continue to live on if one of its elements is
still used because we need to keep track of the transactional heritage.

On Tue, Oct 22, 2013 at 1:52 PM, jkschneider notifications@github.comwrote:

> I know it's hard to believe at first, but I have found that closing the
> transaction does not allow the cache contents to be GC'ed. I could try to
> go back and write a test to prove this out. One of my instances was
> consistently throwing excessive GC overhead exceptions under heavy load
> until this fix was applied. MAT revealed the following:
> 
> [image: image]https://f.cloud.github.com/assets/1697736/1385263/a5063918-3b5b-11e3-8618-904d714177db.png
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/444#issuecomment-26846633
> .

## 

Matthias Broecheler
http://www.matthiasb.com

jkschneider:
Appreciate the feedback, guys.  I'll try to write up a test to demonstrate.

mbroecheler:
Thank you! We are not trying to dismiss your point - just want to make sure
we blame the right part ;-)

On Tue, Oct 22, 2013 at 7:52 PM, jkschneider notifications@github.comwrote:

> Appreciate the feedback, guys. I'll try to write up a test to demonstrate.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/444#issuecomment-26877036
> .

## 

Matthias Broecheler
http://www.matthiasb.com

mbroecheler:
BTW, we have significantly rewritten the caching infrastructure to make it more performant. Those improvements are slated for 0.4.1 and are currently on master. Are you still seeing the memory issue?

jkschneider:
Matthias,

The work you and Pavel did on caching has cleared up the memory condition.  Thanks!

Jon

Issue #510, opened 2013-12-04
gremlin.sh doesn't parse more than 9 parameters


tsecheran:
If more than 9 parameters passed onto gremlin.sh,  gremlin interprets 10 as ${1}0, so the 10th param would result in "-e0", this is due to this following eval in the parsing block

if [ $# -gt 2 ]; then
    for (( i=3 ; i < $# + 1 ; i++ ))
    do
      eval a=\$$i
      k="$k \"$a\""
    done
  fi

**\* eval a=\${${i}} **\* should fix this issue...

dalaro:
I merged your associated PR (#513) into master.  This part of gremlin.sh looks needlessly byzantine... it seems like the loop and eval block could be replaced with a couple of conditional calls to the `shift` builtin.  Still, your fix looks correct and in testing seems to handle both arguments with internal spaces and positional arguments greater than $9.  Thanks.

tsecheran:
Thanks for the merging the fix.

PR #513, opened 2013-12-05
Fix for gremlin.sh will take more than 9 params https://github.com/thinkaurelius/titan/issues/510

tsecheran:
Fix for the issue https://github.com/thinkaurelius/titan/issues/510, please review this and accept the patch.

### DIFF
--- a/titan-dist/src/assembly/static/bin/gremlin.sh
+++ b/titan-dist/src/assembly/static/bin/gremlin.sh
@@ -31,7 +31,7 @@ if [ "$1" = "-e" ]; then
   if [ $# -gt 2 ]; then
     for (( i=3 ; i < $# + 1 ; i++ ))
     do
-      eval a=\$$i
+      eval a=\${${i}}
       k="$k \"$a\""
     done
   fi


dalaro:
Looks good to me.  Gave it a quick test:

``` bash
$ cat test.groovy 
for (a in args) {
    println(a)
}
$ bin/gremlin.sh -e test.groovy `seq 0 20`
0
1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
$ bin/gremlin.sh -e test.groovy "a b" "c"
a b
c
```

spmallette:
Should this change be replicated in Gremlin?

dalaro:
I think so. `${var}` is ancient syntax.  Since gremlin.sh already uses the relatively new `$(command)` substitution syntax, it must be safe to use `${var}`.

spmallette:
Just pushed the same change over in gremlin.  As Titan moves to support TinkerPop 2.5.0, we need to convert the Titan custom Gremlin REPL to a "plugin" for Gremlin.  Then we don't need to maintain this code in two places anymore.  cc/ @mbroecheler

mbroecheler:
Stephen, can you take care of this aspect for the 0.5 release of Titan?

On Thu, Dec 19, 2013 at 6:24 AM, stephen mallette
notifications@github.comwrote:

> Just pushed the same change over in gremlin. As Titan moves to support
> TinkerPop 2.5.0, we need to convert the Titan custom Gremlin REPL to a
> "plugin" for Gremlin. Then we don't need to maintain this code in two
> places anymore. cc/ @mbroecheler https://github.com/mbroecheler
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/513#issuecomment-30931836
> .

## 

Matthias Broecheler
http://www.matthiasb.com

spmallette:
sure...by "we" i basically meant "me"

mbroecheler:
Ah, the royal "we" - got it ;-)

On Thu, Dec 19, 2013 at 4:05 PM, stephen mallette
notifications@github.comwrote:

> sure...by "we" i basically meant "me"
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/thinkaurelius/titan/pull/513#issuecomment-30978847
> .

## 

Matthias Broecheler
http://www.matthiasb.com

PR #518, opened 2013-12-10
If there is only one thread, then the threadpool is null and this errors

mpouttuclarke:


dalaro:
This patch looks good.  fix-for-performance is already merged into master, so I pulled your commit into a local testing branch and then to master: e50826bd9866ca1fc1b8844a18f1dfdc0e122629.

Issue #165, opened 2013-02-27
Titan + HBase hangs at Gremlin console


cglewis:
Running on Ubuntu 12.04 LTS x86_64

Here's my maven version:

```
    user@host:~/titan-0.2.0$ mvn -v
    Apache Maven 3.0.5 (r01de14724cdef164cd33c7c8c2fe155faf9602da; 2013-02-19 13:51:28+0000)
    Maven home: /opt/apache-maven-3.0.5
    Java version: 1.6.0_32, vendor: Sun Microsystems Inc.
    Java home: /usr/java/jdk1.6.0_32/jre
    Default locale: en_US, platform encoding: UTF-8
    OS name: "linux", version: "3.2.0-27-virtual", arch: "amd64", family: "unix"
```

And running configuration for my remote HBase cluster (running CDH4.2 with Hbase 0.94.2-cdh4.2.0 and Titan 0.2.0):

```
    user@host:~/titan-0.2.0$ ./bin/gremlin.sh

             \,,,/
             (o o)
    -----oOOo-(_)-oOOo-----
    gremlin> conf = new BaseConfiguration();                                                                                                                                    
    ==>org.apache.commons.configuration.BaseConfiguration@1dbef987
    gremlin> conf.setProperty("storage.backend","hbase");                                                                                                                       
    ==>null
    gremlin> conf.setProperty("storage.hostname","x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x,x.x.x.x");
    ==>null
    gremlin> TitanGraph g = TitanFactory.open(conf);                                                                                                                            
    13/02/27 19:31:40 INFO zookeeper.ZooKeeper: Client environment:zookeeper.version=3.4.3-1240972, built on 02/06/2012 10:48 GMT
    13/02/27 19:31:40 INFO zookeeper.ZooKeeper: Client environment:host.name=host
    13/02/27 19:31:40 INFO zookeeper.ZooKeeper: Client environment:java.version=1.6.0_32
    13/02/27 19:31:40 INFO zookeeper.ZooKeeper: Client environment:java.vendor=Sun Microsystems Inc.
    13/02/27 19:31:40 INFO zookeeper.ZooKeeper: Client environment:java.home=/usr/java/jdk1.6.0_32/jre
[...]

PR #532, opened 2013-12-18
MapR Tables (HBase) supported as Titan Backend

boorad:
Here are my changes that were required to use MapR jar files.  I think they're unobtrusive, as you only use this with the 'mapr-m7' maven profile.

Also, there was a call to what I would characterize as unreliable modifyTable() in HBase.  This patch works with both MapR tables as well as Apache HBase, and is probably the solution to thinkaurelius/titan#165.

### DIFF
--- a/README.textile
+++ b/README.textile
@@ -11,6 +11,7 @@ h2. Features
 * Support for various "storage backends":https://github.com/thinkaurelius/titan/wiki/Storage-Backend-Overview:
 ** "Apache Cassandra":http://cassandra.apache.org/ (distributed)
 ** "Apache HBase":http://hbase.apache.org/ (distributed)
+** "MapR Tables":http://www.mapr.com/products/mapr-editions/m7-edition (distributed)
 ** "Oracle BerkeleyDB":http://www.oracle.com/technetwork/database/berkeleydb/overview/index-093405.html (local)
 ** "Persistit":https://github.com/pdbeaman/persistit (local)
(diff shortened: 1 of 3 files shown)


dalaro:
This is great, I'm glad you set this up in a pom profile.  I have one question before merging: can we keep the `setCompressionType` call? dbc446e0c2ded75f08f4e52db786c3726576e5d0 deletes it but I'd like to keep it if MapR supports it.

boorad:
We can keep it.  But MapR compresses naively.  LZ4 by default.  If you
like, it can stay for stock hbase.

Cheers,
BA

dalaro:
I added an optional Compression.Algorithm override option in #584.  Leaving it unspecified lets the underlying implementation use its default (to get no compression one would have to explicitly set it to NONE).

I also merged this into titan05 with aca2ec3227ae78308ac3250fcee6f075887c8c48.  I recently split the titan-hbase module into titan-hbase-094 and -096 submodules, and I put the MapR profile in -094.

The MapR profile threw a ClassNotFoundException in testing on a ZK class when I used version 3.3.6 as specified in the PR.  When I used ZK 3.4.5 the suite passed.  Do you know what's going on there?

dalaro:
Changed the ZK version used in the MapR build to 3.4.5 in 061e780f650e7cd8d57329879d914b44447365b6.  Please reopen if there's something I'm missing.

Issue #526, opened 2013-12-17
Titan on HDP2, HBase 0.96


ghost:
Hi, I'm developing a solution on the Hortonworks Data Platform. This issue is about Titan-HBase. It seems like the existing repo only supports HBase 0.94. There are some API changes in 0.96, which is shipped with the Hortonworks HDP2 GA. Here's the error that it will currently produce:

```
[root@sandbox titan-hbase-0.4.1]# ./bin/gremlin.sh 
[...]
gremlin> conf = new BaseConfiguration();
==>org.apache.commons.configuration.BaseConfiguration@c8282a6
gremlin> conf.setProperty("storage.backend","hbase");
==>null
gremlin> conf.setProperty("storage.hostname","sandbox.hortonworks.com");
==>null
gremlin> g = TitanFactory.open(conf);
Not a host:port pair: PBUF
$
sandbox.hortonworks.com?????(
Display stack trace? [yN] 
```

It doesn't seem like it's much more than that that they've restructured the packages in regard to the use that Titan has of it. To make it work I did a couple of changes in HBaseStoreManager.java and the pom:

HBaseStoreManager.java
Line 23: `import org.apache.hadoop.hbase.io.compress.Compression;`
Line 402: `commands.setSecond(new Delete(key.as(StaticBuffer.ARRAY_FACTORY), (int) delTimestamp, 0));`

titan-hbase/pom.xml
Line 13: `<hbase.version>0.96.0-hadoop2</hbase.version>` (may be that you can set that at command line as well.
Added dependencies:
    <groupId>org.apache.hbase</groupId>
        <artifactId>hbase-common</artifactId>
        <version>0.96.0-hadoop2</version>
    </dependency>   

Line 46: `<artifactId>hbase-client</artifactId>`

Then I compiled everything as mentioned in the HBase section of the documentation here at Github: `mvn clean install -Dhadoop.profile=2.0 -Dhadoop.version=2.2.0 -DskipTests=true`

```
gremlin> conf = new BaseConfiguration();
==>org.apache.commons.configuration.BaseConfiguration@50f47076
gremlin> conf.setProperty("storage.backend","hbase");
==>null
gremlin> conf.setProperty("storage.hostname","sandbox.hortonworks.com");
==>null
gremlin> conf.setProperty("storage.port","2181");
==>null
gremlin> g = TitanFactory.open(conf);
[...]

PR #539, opened 2013-12-20
Changed APIs and pom to HBase 0.96-hadoop2. Breaks compability with 0.94...

ghost:
Regarding issue #526.

### DIFF
--- a/titan-hbase/pom.xml
+++ b/titan-hbase/pom.xml
@@ -10,8 +10,8 @@
     <name>Titan-HBase: Distributed Graph Database</name>
     <url>http://thinkaurelius.github.com/titan/</url>
     <properties>
-        <hbase.version>0.94.12</hbase.version>
-        <hadoop.version>1.0.4</hadoop.version>
+        <hbase.version>0.96.0-hadoop2</hbase.version>
+        <hadoop.version>2.2.0</hadoop.version>
         <hadoop.artifact>hadoop-core</hadoop.artifact>
         <titan.hbase.mvn.opts />
         <default.test.jvm.opts>-Xms256m -Xmx1280m -XX:+HeapDumpOnOutOfMemoryError -ea ${test.extra.jvm.opts}</default.test.jvm.opts>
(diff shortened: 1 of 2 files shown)


ghost:
Regarding issue #526.

Omitted the row size in the Delete call in order to handle the deletion issues with the previous pull request.

Tested and verified that it now successfully adds and deletes edges and vertices up until HBase 0.96.1.1.

dalaro:
Thanks for putting this together.

You're right to fix up that Delete call.  I want to remove all the deprecated calls from the storage adapter at some point.

We're not ready to break Hadoop 1 and HBase 0.94 compatibility yet, so I think we'll have to implement this differently.

I have a specific question about HDP vs Apache client artifacts.  Does Titan need to use the HBase and Hadoop artifacts hosted in Hortonworks's repo @ http://repo.hortonworks.com/content/repositories/releases/org/apache/ to be compatible with HDP2 clusters, or do Apache client artifacts of similar versions work just as well when talking to HDP services?

dalaro:
I'm refactoring the titan-hbase module into separate 0.96 and 0.94 builds.  The vast majority of the code and all of the tests still reside in a single set of files in titan-hbase/src.  I moved the one 0.94 <-> 0.96 source incompatibility, Compression.Algorithm moving between packages, into a little helper class which is defined separately for each build.

The default for 0.96 is to build against 0.96.1.1-hadoop2 with the default Hadoop version (2.2.0). 

I'm probably going to tweak this some more, but the changes I just described are in titan05 now.  I like this approach because it lets us keep compatibility with multiple HBase + Hadoop stack versions.

ghost:
This is great news. 

Re compability between HDP2 and the Apache artifacts, I've been using the Apache ones with titan05 in my sandbox. I know they have some special versions in there, but it's been working quite well for me at least. Probably someone from Hortonworks can answer that definitively.

dalaro:
I think the changes from this PR have all made it into titan05 in slightly different forms.  The Delete constructor call got refactored out of existence as part of configurable timestamp support in f0d41c6743a3514966326d90e9b1234c5d58e16e.  Hadoop 2.x / HBase 0.96 modules and Maven profiles are also in titan05.

So I'm closing this, but please reopen if I missed something.

Issue #578, opened 2014-02-03
Connection pool does not work properly under heavy load.


haebin:
Under heavy load, Titan produces too many TIME_WAIT state sockets.

dalaro:
I think I ran into the same behavior mentioned here, in #582, and #579 while working on testing titan-hadoop.  Setting minIdle=0, maxactive=-1, and maxTotal=positive number seems to cause commons-pool to invoke `clearOldest` and destroy 15% of the total pool capacity every time it fills, so that the pool periodically turns over about a sixth of its capacity when it grows to capacity.

I refactored Titan's use of commons-pool in 7481943407fbbb3ee4e664a1e7e25b25a0d85107, making all of the pool parameters configurable.  I also changed the default settings in 88a06b1045d0d3e2bc047a46b1002a50b5059fc1, so that it now uses maxtotal=-1, maxactive=16, maxidle=4 by default.  This seems to avoid continual `clearOldest` invocations in normal operation.  These commits are part of 0.5.1.

This change solved TIME_WAIT socket generation in titan-hadoop testing, where test MapReduce jobs backed by Cassandra were generating thousands of lingering sockets.  The new settings produced an order of magnitude reduction.  So I'm hoping this also addresses the problem that you saw.  Please reopen if this still occurs.

PR #579, opened 2014-02-03
Connection pool fixed since it did not work as advertised. Under heavy

haebin:
load, old connection pool closes and opens connections to Cassandra too
often generating too many TIME_WAIT state sockets which eat up all the
ephemeral ports. Since connections are local, let there be idle (warm)
connections.

https://github.com/thinkaurelius/titan/issues/578


haebin:
Due to warmup code, Titan cannot startup when there is no scheme setup.
Sorry.

Issue #580, opened 2014-02-03
Needs per transaction read/write consistency level setting.


haebin:
Needs the option for it.

dalaro:
Please see #580

PR #581, opened 2014-02-03
Per transaction consistency setting feature added. Can be used liked the

haebin:
following.

g.getCurrentThreadTx().getTxHandle().getStoreTransactionHandle().getBaseTransaction().getWrappedTransactionHandle().setReadConsistencyLevel("ONE");

Ugly to use, yet less intrusive to implement the feature.

https://github.com/thinkaurelius/titan/issues/580


dalaro:
This enough of a feature that I'm more comfortable with it in titan05 than master, but it's a fuzzy judgment.  titan05 has recently seen lots of refactoring around configuration options.  titan05 supports transaction-specific consistency level overrides like so:

```
t = g.buildTransaction().setCustomOption('storage.read-consistency-level', 'ONE').start()
```

There is also a `storage.write-consistency-level` option.

We may change the API again before 0.5 releases.

PR #582, opened 2014-02-05
Connection pool fixed since it did not work as intended. Under heavy …

haebin:
load, old connection pool closes and opens connections to Cassandra too
often generating too many TIME_WAIT state sockets which eat up all the
ephemeral ports. Since connections are local, let there be idle (warm)
connections.

https://github.com/thinkaurelius/titan/issues/578

### DIFF
--- a/titan-cassandra/src/main/java/com/thinkaurelius/titan/diskstorage/cassandra/thrift/CassandraThriftStoreManager.java
+++ b/titan-cassandra/src/main/java/com/thinkaurelius/titan/diskstorage/cassandra/thrift/CassandraThriftStoreManager.java
@@ -89,13 +89,17 @@ public class CassandraThriftStoreManager extends AbstractCassandraStoreManager {
         p.setTestOnBorrow(true);
         p.setTestOnReturn(true);
         p.setTestWhileIdle(false);
-        p.setWhenExhaustedAction(GenericKeyedObjectPool.WHEN_EXHAUSTED_BLOCK);
-        p.setMaxActive(-1); // "A negative value indicates no limit"
-        p.setMaxTotal(maxTotalConnections); // maxTotal limits active + idle
-        p.setMinIdle(0); // prevent evictor from eagerly creating unused connections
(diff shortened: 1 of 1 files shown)


dalaro:
@haebin thanks for this.  Can you help me understand the problem a bit better?

How many TIME_WAIT connections are you seeing in `netstat`?  What did you have `connection-pool-size` set to?  Is the box running one graph or multiple graphs?  How long does a graph instance live between creation and `shutdown()`?

The default pool size is 32 total with max 8 idle and a 60 second idle grace period, so the evictor should kill no more than about 24 a minute (per graph/manager instance).  Assuming TCP MSL of 60 sec and 2 \* MSL wait time, I think there's a ceiling of about 48 lingering TIME_WAIT sockets in the pathological case for a single graph at steady state.  So I'm surprised that you are seeing ephemeral port exhaustion -- what am I missing?

haebin:
It's Titan/Cassandra (0.4.1) 
I tried many options for connection pool size, such as, 32, 64 and such.
Per request it takes around 1000ms. (It pulls a lot of data from some other systems and performs recommendation based on that.)

If you try to run the heavy duty query test by grinder (or nGrinder) with virtual user greater than 40 then Titan produces many error outputs with Thrift connection exception. Something like 'cannot create sockets' or something.

I will try to reproduce the error and put up the error message and TIME_WAIT count.

haebin:
Hmm, I was just too busy to follow up. Let me just summarize what I have investigated. Due to commons pool implementation, the previous configuration opens and closes connections too often under heavy load. Since this is a connection pool, I don't see a reason why it needs to evict any connections unless it is broken. Leaving connections idle would not cause any harm especially for the setup where Titan and Cassandra are on the same machine. With the above modification, my Titan node was able to withstand more than 40 VUser with very very heavy query. Now, it can withstand 100 VUser yet the Mean Transaction Time increased by double. Before I made this change, the Titan node was just too busy producing errors with more than 40 VUser.

dalaro:
See #578

PR #602, opened 2014-02-25
Allow Geoshapes to be parsed from a variety of formats

### LINKED ISSUE #572: GeoshapeHandler should work with the type 'List'
`GeoshapeHandler` can parse a `Geoshape` out from a couple types: `float[]`, `double[]`, and a `String with the format`lat,long`or`lat;long`. Unfortunately it cannot handle parsing a`List`type. Adding support for this would allow the`GraphSONReader` to take this:

```
{
    "vertices": [
      {"_id": 0, "location": [45, 46]},
    ]
}
```

And treat the `"location"` field as a `Geoshape`.
 a `double

erickt:
This allows Geoshapes to be extract from a java `List`, or parsed from `"point[123,456]"`, `"circle[123,456,789]"`, and `"box[12,34,56,78]"`.

Fixes #572


erickt:
Good morning. Has there been any discussion on if this is the right approach for implementing this?

mbroecheler:
This has been implemented differently using JSON.

PR #605, opened 2014-03-05
Use elasticsearch 1.0.0

avieth:
Changelog
- Use elasticsearch 1.0.0 and lucene 4.6.1; to be determined whether the lucene version bump breaks anything.
- Change `SearchRequestBuilder.setFilter` to `SearchRequestBuilder.setPostFilter` [see this](http://www.elasticsearch.org/guide/en/elasticsearch/reference/1.x/_search_requests.html).

Are there any other changes that must be made?

`mvn test` passes on my machine.

### DIFF
--- a/titan-es/pom.xml
+++ b/titan-es/pom.xml
@@ -19,7 +19,7 @@
           which it depends and modify ../titan-lucene/pom.xml
           as necessary to keep the versions matched.
         -->
-        <elasticsearch.version>0.90.5</elasticsearch.version>
+        <elasticsearch.version>1.0.0</elasticsearch.version>
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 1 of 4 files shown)


dhardy92:
Maybe you should update to latest 1.0.1 ?

avieth:
> Maybe you should update to latest 1.0.1 ?

I presume that there are no breaking changes between ES 1.0.0 and 1.0.1, so this titan branch built against ES 1.0.0 should work with ES 1.0.1.

dalaro:
I merged this into titan05 in a13fc958fbb7a9a5f04d2a760d7c1bb1304cd14c and updated to 1.0.1 in 0385d74d3f8ebbbfa5887b1a8c7a0aa8d3701645.

We could backport this to master (0.4.x) in a separate issue.  If we backport to 0.4.x, this would need an option to switch between 0.90.x and 1.0 (or autodetect logic to switch) so that users who don't want to upgrade ES yet don't have to.

mbroecheler:
No, let's not port this back to the 0.4 line since this would cause a major incompatibility. 0.5 is close enough ;-)

dalaro:
Sounds good, thanks.

PR #609, opened 2014-03-13
titan 0.4.2 hadoop 2.2 support

erickt:
This PR backports @dalaro's hadoop 2.2 support back onto 0.4.2. I don't necessarily expect this PR to be landed as is though. If this is worthwhile, I would suggest instead landing it as a separate branch so people can upgrade to hadoop 2.2 distributions and continue to use Titan 0.4.2 in production while 0.5.0 is in development.


mbroecheler:
Thanks @erickt. We will leave this PR open for people to grab it but we won't support it for the 0.4.x line of Titan but only in 0.5.x and onward.

PR #620, opened 2014-04-02
Remove need for admin/create permissions in HBase prod tables

### LINKED ISSUE #619: Titan requires create/admin access for HBase tables
Titan does a bunch of checks to ensure the table and column family exists. All these require admin/create access. These checks are not needed in production machines, where tables are already created once.

We can disable this check on production.

Janarthanam:
fix #619


dalaro:
I merged this to master, but renamed the option to "skip-schema-check".  I also ported it forward to titan05.

PR #632, opened 2014-04-22
must contains 0.4.2

boliza:
if not contains 0.4.2,can't open the titan instance which is create by titan 0.4.2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
         <tag>HEAD</tag>
     </scm>
     <properties>
-        <titan.compatible.versions>0.4.0,0.4.1</titan.compatible.versions>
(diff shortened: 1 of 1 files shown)


dalaro:
Whoops.  This is severe.  Work on it now.

boliza:
could I redownload to fixed it?

dalaro:
I built Titan 0.4.4 zipfiles that are compatible with 0.4.2 and 0.4.3:

https://github.com/thinkaurelius/titan/wiki/Downloads#titan-044

Artifacts are forthcoming to Maven Central later today (they aren't synced yet).

boliza:
thanks for your work

PR #633, opened 2014-04-22
Add test for unregistering a log reader.

solongordon:
Here's a small commit to try out the Knewton↔Aurelius workflow. I added a unit test for the Log.unregisterReader(...) method.

Update: I added a commit to generally speed up LogTest. It uses CountDownLatches rather than sleeps, which avoids unnecessary waiting.


mbroecheler:
Looks good. I like the use of CountDownLatches to speed things up.

Issue #651, opened 2014-05-28
Using "Contain" predicate on sorted key breaks multi vertex queries


mgmuscari:
There is a corner case in the AbstractVertexCentricQueryBuilder class:

java.lang.ClassCastException: com.thinkaurelius.titan.graphdb.query.condition.Or cannot be cast to com.thinkaurelius.titan.graphdb.query.condition.PredicateCondition at com.thinkaurelius.titan.graphdb.query.AbstractVertexCentricQueryBuilder.compileSortKeyConstraints(AbstractVertexCentricQueryBuilder.java:342)...

There needs to be a 2-stage check here that first verifies that the Condition is a PredicateCondition before attempting to cast. Note that this is done in the block of code following the affected one.

mgmuscari:
A link to the block of code in question:

https://github.com/thinkaurelius/titan/blob/master/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/query/AbstractVertexCentricQueryBuilder.java#L342

PR #653, opened 2014-05-29
Fix ClassCastException in AbstractVertexCentricQueryBuilder

mgmuscari:
Addresses https://github.com/thinkaurelius/titan/issues/651

Add a check for Conditions to ensure they can be casted to PredicateCondition
before attempting the cast while looking for equality constraints in the
AbstractVertexCentricQueryBuilder


mgmuscari:
i'm pretty sure the travis CI build failed due to the test suite taking a long time to complete...

mbroecheler:
yes, unfortunately it tends to do that. Seems Travis wasn't designed to build database software with long and expensive test cases ;-)
Thanks, we'll take a look.

mbroecheler:
This does not need to be merged upstream since it has already been solved in titan 05.

PR #663, opened 2014-06-14
bin/ scripts to follow symlinks

brennonyork:
Since many applications now symlink one, or more, times into multiple directories these scripts needed updated to accurately follow said symlinks. Previously, if symlinks were present, the script would fail to find all of the .jar files in the library because it would only find the immediate path it was installed in. For example, if we symlinked rexster.sh to /usr/local/bin/rexster and attempted to call it, the script would only look in /usr/local/lib/ for its jar files which would fail to launch.

This is similar to the patch for the rexster-console script that was submitted a while back (tinkerpop/rexster#353) which details the discussion and problem.

The impetus for this patch is to get it fixed and in the upstream version for the next release build such that the homebrew formula to install titan (homebrew/homebrew#30107) can be accepted for the wider community.

Let me know if there are any issues or concerns!

### DIFF
--- a/titan-dist/src/assembly/static/bin/cassandra
+++ b/titan-dist/src/assembly/static/bin/cassandra
@@ -36,7 +36,20 @@
 # NB: Developers should be aware that this script should remain compatible with
 # POSIX sh and Solaris sh. This means, in particular, no $(( )) and no $( ).
 
-export CASSANDRA_INCLUDE="`dirname $0`/cassandra.in.sh"
+# Returns the absolute path of this script regardless of symlinks
+abs_path() {
+    # From: http://stackoverflow.com/a/246128
+    #   - To resolve finding the directory after symlinks
+    SOURCE="${BASH_SOURCE[0]}"
+    while [ -h "$SOURCE" ]; do
+        DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+        SOURCE="$(readlink "$SOURCE")"
+        [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
+    done
+    echo "$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+}
+
+export CASSANDRA_INCLUDE="`abs_path`/cassandra.in.sh"
 . "$CASSANDRA_INCLUDE"
 
 # Use JAVA_HOME if set, otherwise look for java in PATH
--- a/titan-dist/src/assembly/static/bin/cassandra.in.sh
+++ b/titan-dist/src/assembly/static/bin/cassandra.in.sh
@@ -27,7 +27,7 @@ abs_path() {
     echo "$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 }
 
-CASSANDRA_HOME=$(cd `abs_path` && pwd)
+CASSANDRA_HOME=$(cd `abs_path`/.. && pwd)
 
 # The directory where Cassandra's configs live (required)
 CASSANDRA_CONF=$CASSANDRA_HOME/conf
--- a/titan-dist/src/assembly/static/bin/cassandra.in.sh
+++ b/titan-dist/src/assembly/static/bin/cassandra.in.sh
@@ -14,8 +14,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cd `dirname $0`/..
-CASSANDRA_HOME=`pwd`
+# Returns the absolute path of this script regardless of symlinks
+abs_path() {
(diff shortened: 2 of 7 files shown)


brennonyork:
@dalaro @spmallette @joshsh Can any of you assist with understanding why this build is failing with the following errors from the Travis CI server?

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
157  [pool-1-thread-2] WARN  org.apache.cassandra.utils.CLibrary  - Unable to lock JVM memory (ENOMEM). This can result in part of the JVM being swapped out, especially with mmapped I/O enabled. Increase RLIMIT_MEMLOCK or run Cassandra as root.
3987 [pool-1-thread-2] WARN  org.apache.cassandra.db.SystemTable  - No host ID found, created 23d7406f-48b5-4afb-93da-58602bfb16b4 (Note: This should happen exactly once per node).
11763 [COMMIT-LOG-WRITER] ERROR org.apache.cassandra.db.commitlog.CommitLogSegment  - Attempted to write commit log entry for unrecognized column family: eb9a04e4-8e46-3d8b-86b3-fea561e4e47a
24621 [MemoryMeter:1] WARN  org.apache.cassandra.db.Memtable  - setting live ratio to maximum of 64.0 instead of 76.69594594594595
25967 [MemoryMeter:1] WARN  org.apache.cassandra.db.Memtable  - setting live ratio to maximum of 64.0 instead of 78.72297297297297
64574 [MemoryMeter:1] WARN  org.apache.cassandra.db.Memtable  - setting live ratio to maximum of 64.0 instead of 64.14154926604084
65127 [MemoryMeter:1] WARN  org.apache.cassandra.db.Memtable  - setting live ratio to maximum of 64.0 instead of 64.14154926604084
65671 [MemoryMeter:1] WARN  org.apache.cassandra.db.Memtable  - setting live ratio to maximum of 64.0 instead of 64.14154926604084
70904 [MemoryMeter:1] WARN  org.apache.cassandra.db.Memtable  - setting live ratio to maximum of 64.0 instead of 73.86497326203208
70962 [MemoryMeter:1] WARN  org.apache.cassandra.db.Memtable  - setting live ratio to maximum of 64.0 instead of 73.86497326203208
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

I attempted to add and remove an empty file (the last two commits) to trigger another Travis build under the assumption that there might have been something up with the CI servers on the previous run, but still the same failure.

I've been able to run the tests locally with `mvn test -B` and all succeeded. I did notice that once it started the test suite for **titan-cassandra** my CPU pegged out at ~750% for the java processes (at 8 cores). I'm wondering if the Travis CI system can handle that sort of CPU necessity or if the scheduling for compute resources on the Travis system create an issue where the titan tests cannot execute in a quick enough manner. That said, the ENOMEM warning could mean the Travis build system just can't allocate enough memory for the tests to execute. Regardless, any help in this would be greatly appreciated as I'm completely befuddled in how adding symlinks to the deploy scripts would cause a major crash of the titan system, especially when it still builds and tests fine locally.

dalaro:
@brennonyork  Those Travis failures are spurious.  Sorry for the confusion. Travis's 50 minute walltime limit has stymied us for a while; we keep running into the 50 minute limit.  I'm working on a CI alternative for Titan.  I tried to disable Travis builds by toggling the "Build only if .travis.yml is present" switch in our settings to "ON" and deleting .travis.yml, since this has led to confusion on PRs at least twice in recent memory.  Not sure if that will actually disable Travis, but we'll find out shortly...

brennonyork:
@dalaro So is there anything you recommend for me to do to test and ensure this pull can be successfully merged into the main branch? I'm hoping to get this resolved so I can get Titan added into the Homebrew community. Any advice would be great, thanks!

mbroecheler:
@brennonyork We are in the process of moving to bamboo. Until then, please build locally.

dalaro:
Thanks for contributing this.  That bash wart is not exactly graceful compared to `readlink -f`, but `readlink -f` doesn't work on OSX and the wart does (should? I haven't actually tried).  So the wart wins.

We're close to a 0.5.0 release, and I'd like to close this for the milestone, but I also don't want to merge master into titan05 this late.  So I rebased your changes onto titan05 and bypassed master.

Specifically, I branched titan05 to titan05-readlink, then ran `git rebase --ignore-date --onto titan05-readlink brennonyork/master brennonyork/symlink-scripts`.  There was a minor conflict in gremlin.sh.  I also propagated the readlink blob to a couple of other scripts added in titan05 that didn't exist on master and hence weren't in the PR.

PR #665, opened 2014-06-20
Add failing test for edges exceeding tx-cache-size

solongordon:
There seems to be buggy behavior when the number of edges on a vertex
exceeds the tx-cache-size number. If you remove an edge from the vertex,
it can reappear when you call getEdges on the vertex. I added a unit
test which demonstrates this.


dalaro:
Your test reproduced the problem on my machine.  I'll cherry pick the test forward to titan05 and open an issue to fix this in both branches (assuming it applies to titan05 and not just master).  This is some pathological behavior.  Thanks for isolating it and commenting the test beautifully.

dalaro:
Cherry-picked to titan05 as a31802d0fd2cb8b31ebdb8f6173f9b7a876a6f40.  It didn't fail when I ran it, but if the root cause turns out to be nondeterministic then that means nothing.

#668 remains open to fix the bug

PR #692, opened 2014-07-28
FoundationDB storage backend

MMcM:
Mostly just new files. All dependencies should be in Maven Central.


dalaro:
Thanks for turning this into a PR.  I branched off titan05 @ ce34565218b3c8ce63bc33b69f26ea3ddc487c6d as fdb05, then merged.  I saw your update and then deleted fdb05 and repeated the process, since I got conflicts somehow when I attempted to pull the updated McMM:fdb-0.5 onto the original thinkaurelius:fdb05.  Not sure what I did wrong there.

Aside from installing FDB and starting the server, do I need to do any setup before running the tests?

mbroecheler:
FoundationDB no longer exists.

Issue #740, opened 2014-09-01
Outdated config elements in GraphOfTheGodsFactory.create


dalaro:
https://github.com/thinkaurelius/titan/blob/0.5.0/titan-core/src/main/java/com/thinkaurelius/titan/example/GraphOfTheGodsFactory.java#L28-L40

Running `g = TitanFactory.open('path/to/cfg'); GraphOfTheGodsFactory.load(g)` is the workaround.

tomersagi:
Could you give an example of TitanFactory.open('path/to/cfg')?
the following returned an error:

> gremlin> g = TitanFactory.open('../conf/titan-berkeleydb.properties'); GraphOfTheGodsFactory.load(g)
> Backend shorthand unknown: ../conf/titan-berkeleydb.properties

dalaro:
Try this:
1. unzip titan 0.5.0's zipfile
2. cd into the titan-0.5.0-hadoop1 or -hadoop2 directory (according to whichever version you downloaded and unzipped)
3. run `bin/gremlin.sh`
4. type at the gremlin prompt as follows (assuming the path separator char on your platform is `/`):

```
gremlin> g = TitanFactory.open('conf/titan-berkeleydb-es.properties')
==>titangraph[berkeleyje:/home/dalaro/tinkerelius/titan-0.5.0-hadoop2/conf/../db/berkeley]
gremlin> GraphOfTheGodsFactory.load(g)
==>null
gremlin> g.V.map
03:39:16 WARN  com.thinkaurelius.titan.graphdb.transaction.StandardTitanTx  - Query requires iterating over all vertices [()]. For better performance, use indexes
==>{name=saturn, age=10000}
==>{name=sky}
==>{name=sea}
==>{name=jupiter, age=5000}
==>{name=neptune, age=4500}
==>{name=hercules, age=30}
==>{name=alcmene, age=45}
==>{name=pluto, age=4000}
==>{name=nemean}
==>{name=hydra}
==>{name=cerberus}
==>{name=tartarus}
gremlin>
```

The `titan-berkeleydb.properties` file that ships with Titan is not useful for GotG because it only configures a storage backend without an index.  Note that I used `titan-berkeleydb-es.properties` above, which does configure an index.

tomersagi:
Great, works fine now. 
Thanks

ccoffey:
[...]

PR #760, opened 2014-09-16
patch docs for 0.5.x tutorial based on #740

boorad:
I hit these issues while looking at the docs in s3, so thought I'd change them around a bit.  Not sure if you want to go with berkeleydb-es as the most basic conf file for the intro tutorial, but I'm not sure if local storage still exists.  Also, I don't know if ../db has to exist before running this, as I had already modified the .properties file to other paths.

### DIFF
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -171,7 +171,7 @@ Archive:  titan.zip
   creating: titan/
 ...
 $ cd titan
-$ bin/gremlin.sh 
+$ bin/gremlin.sh
 
          \,,,/
          (o o)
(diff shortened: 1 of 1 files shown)


dalaro:
Thanks.  The "local" storage backend name is indeed gone in 0.5.  I merged your changes, then rewrote the paragraph above your code edit that mentioned the "local" anachronism in deb2353750c6e29ff0d7f27730afdd13d6d20b0e.  Please let me know if you spotted other uses of "local" that I missed.  I updated the 0.5.0 docs with these two commits: http://s3.thinkaurelius.com/docs/titan/0.5.0/getting-started.html#_loading_the_graph_of_the_gods_into_titan

I'm not sure about the problem with ../db existence for BDB + ES.  It need not exist beforehand (or there's a bug).  Did we write otherwise somewhere?

PR #771, opened 2014-09-30
fix issue  #769

boliza:
@mbroecheler   about a month ago I had modify titan to support Boolean Type.I used this feature in my product about two weeks.
Disscuss here:https://groups.google.com/forum/#!topic/aureliusgraphs/YAbKt9Vscl0
an early pull request I have closed:https://github.com/thinkaurelius/titan/pull/745


boliza:
In lucene.use String to support Boolean type,use Long to support Date type.

PR #792, opened 2014-10-23
pull request1

joisonwk:
add Dockerfile


dalaro:
The lone `RUN` command in this Dockerfile appears to just echo "this is a docker build configfile".  Does this Dockerfile actually do anything?  It looks to me like it does nothing.  Please reopen if you would like to discuss further.

Issue #801, opened 2014-11-02
Could be an error in bdb.html


angelcervera:
Could be an error in titan-config-ref.html

In the documentation :

TitanGraph g = TitanFactory.build()
    .set("storage.backend", "berkeleyje");
    .set("storage.directory", "/tmp/graph");
    .open()

and should be :

TitanGraph g = TitanFactory.build()
    .set("storage.backend", "berkeleyje")
    .set("storage.directory", "/tmp/graph")
    .open();

Pay attention to the ";" at the end of each line.

dalaro:
This is fixed in titan05.  I also cherry-picked this back to 0.5.0_docs and 0.5.1_docs branches that I've been keeping locally off their respective release tags and rebuilt+redeployed the docs for both of those versions.

PR #830, opened 2014-11-14
Docs: fixed syntax in code example

dminkovsky:


dalaro:
I think this is the same as #801.  db0984d640fc1bff7f6cd1a6a1838050eca31d9d includes this change and a couple of other related semicolon-placement inconsistencies in the docs.

PR #852, opened 2014-11-27
remove never uesed temporary variables

boliza:
remove never uesed temporary variables


boliza:
I think has aready fixed

Issue #854, opened 2014-12-01
Upgrade Guava


mbroecheler:
to current version. Remove usage of StopWatch

dalaro:
We should probably shade Titan's copy of Guava.  Some of Titan's dependencies may still be incompatible with versions later than 15.0, since Guava periodically removes deprecated API bits.  Titan's own classes can use the latest version regardless (as long as we shade).

boliza:
Shade Guava would be a good choice.

tomersagi:
+1 for this. This is causing issues downstream when embedding titan with projects using guava 18.

isaki:
+1 for this.

Is it possible to get an ETA for this or a targeted release version?

saden1:
Ran into this issue...Titan should shade all the third party libraries it uses so users aren't prevented from using their own versions of these libraries in their application.

boliza:
titan-core provide some interface need Guava class as param.so shade guava is not a good choice.

saden1:
I don't understand the previous comment.  Shading is the only good choice unless you want users to be indefinitely bound to titan's version of guava or think it's desirable to run into this kind of issues. The best course titan can employ is to either shade it's third party dependencies along with it's jar files or pre-shade and take ownership of third party libraries and use the shaded version of these libraries through out its code.

I think the latter choice is the safest and most flexible choice.

dalaro:
As of cfc37e3195066c3298a8719aa66a946fb749434d, titan09 is on Guava 18 (but not shading Guava).

Upgrading is a step forward.  Shading the Guava classes used the titan-\* modules would be another step forward.  I'm leaving this issue open to see if I can wedge the second step in by 1.0.

dalaro:
After discussion with @mbroecheler, I'm moving shading to 1.1: #1139

PR #870, opened 2014-12-04
use Stopwatch.createStarted() instead of new Stopwatch().start

boliza:
new Stopwatch().start is deprecated. if some project has a guava version above 17, can be an error


boliza:
change to new api also hav a problem in hadoop.close it.
but I think shade guava is necessary。some in issue #854

Issue #574, opened 2014-01-31
Consider GeoshapeHandler being able to parse a GeoJSON


erickt:
[GeoJSON](http://geojson.org/) is a standardized format that is used by a variety of projects, like MongoDB, PostGIS, and Twitter. I don't think it'll be that hard to add support for it in Titan. The approach I see is that the `GeoshapeHandler.convert` could be modified to check if it was passed a `Map<String, Object>` and check if it conforms to the GeoJSON spec.

If we do go this route, I would also suggest either adding a `Map<String, Object> Geoshap.toGeoJSON()` method so that we can export a vertex to JSON and import it back in without everyone needing to write their own GeoJSON formatters and parsers.

on the `Geoshape` itself so that we c

BrynCooke:
These methods could be added to Geoshape:

```
String toGeoJSON()
static Geoshape fromGeoJSON(String json)
```

One potential issue is that geoJSON doesn't support circles. So what do we do if we are trying to serialize one of these? I would not be keen on an approximation via polygon as accuracy requirements may not be known.

There is a proposal for circles in V 1.1 of the spec
https://github.com/geojson/geojson-spec/wiki/Proposal---Circles-and-Ellipses-Geoms

Are any of the other projects using this?

BrynCooke:
Actually the extra methods are an awful idea. Marshalling belongs in type handlers in gson or the equivalent in Jackson.

mbroecheler:
@BrynCooke Exactly, you can extend the TypeHandler to look at strings and detect the encoding, then try to parse.

BrynCooke:
I've not added a toGeoJSON method as I believe that serialization should be performed by a type serializer.
I think this could be added as part of https://github.com/thinkaurelius/titan/issues/954 and if folks want to reuse it in their own code then they can.

PR #952, opened 2015-01-30
Added GeoJson support

BrynCooke:
https://github.com/thinkaurelius/titan/issues/574

Supports Point, Circle and Polygon (Boxes only)

Also a nice side effect is we can embed GeoJson in GraphSON e.g.

``` json
{
    "variables": {},
    "vertices": [{
        "id": 256,
        "label": "test",
        "type": "vertex",
        "properties": {
            "test": [{
                "id": "jk-74-35x",
                "value": {
                    "type": "Feature",
                    "geometry": {
                        "type": "Point",
                        "coordinates": [125.6, 10.1]
                    },
                    "properties": {
                        "name": "Dinagat Islands"
                    }
                },
                "properties": {}
            }]
        }
    }],
    "edges": []
}
```

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/core/attribute/Geoshape.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/core/attribute/Geoshape.java
@@ -11,6 +11,9 @@ import com.thinkaurelius.titan.graphdb.database.idhandling.VariableLong;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 
 import java.lang.reflect.Array;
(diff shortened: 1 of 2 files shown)


BrynCooke:
Actually I have realised that the 'feature' is optional, so I will tweak and make another pull request.

BrynCooke:
OK I think it's good to merge now

dalaro:
This looks good to me.  The `convert` modifications look self-contained, and the new tests also look good.  The tests passed post-merge here before I pushed.

Issue #863, opened 2014-12-02
Remove oldfulgora package and contents


mbroecheler:
also remove associated unit tests. This should all be GraphComputer from now on.

dalaro:
As we discussed recently, it turns out that we don't actually want to remove the associated unit tests (as would have been done by #966).  We want to instead port the unit tests forward to the new fulgora interfaces.  I removed oldfulgora but kept the tests in 5b8bcce7f2d645fe3e27a93a18ab5ca34bf86de5.  The tests compile as of that commit, though two of them have problems:
- The pagerank test is `@Ignore`d.  Per offline discussion, we're not going to use the TP3 PR impl, but instead a vertexprogram (could be pagerank, could be something else, just as long as it has an interesting access pattern for testing) written and maintained as part of Titan.  I made #989 for this.
- The TitanPartitionGraphTest of degree counting over partitioned vertices is broken.  This issue seems to be independent of the migration from old to new fulgora.  This method is not `@Ignore`d like the pagerank test, so it still breaks builds.

PR #966, opened 2015-02-11
Removing oldfulgora and associated test methods

dalaro:
(#863) This PR deletes:
- the entire oldfulgora package
- the subset of OLAPTest methods that used oldfulgora (but not the remaining methods, which seem to be mostly scanjob related)
- the end of the method `TitanPartitionGraphTest.testVertexPartitioning`, which used oldfulgora

I'm submitting this as a PR for review instead of a direct push in case we want to try to adapt the deleted test methods rather than just deleting them.


dalaro:
Rejecting this PR

We want to refactor the tests that currently depend on oldfulgora so that they use GraphComputer instead

Issue #887, opened 2014-12-13
UUID has no default serializer


holub:
Add default serializer for java.util.UUID class.

PR #970, opened 2015-02-13
Add UUID support

BrynCooke:
https://github.com/thinkaurelius/titan/issues/887
This request adds a UUID serializer.
Strings and Java UUID objects are supported allowing roundtrip via GSON.
Internally UUIDs are stored as two longs.


mikedias:
Hi @BrynCooke, 

This PR should consider the data type register in ElasticSearchIndex, isn't it?

BrynCooke:
@dalaro Should I add mixed index support at the same time? I don't think this will be too difficult.

mbroecheler:
@BrynCooke Sure, why not ;-)

boliza:
@BrynCooke so also fix #769 :)

PR #975, opened 2015-03-02
Support Boolean, Date/Time data types in mixed indexes

BrynCooke:
https://github.com/thinkaurelius/titan/issues/769
Added support in ES, Solr and Lucene.


boliza:
LGTM!
I think if push to titan05 will be great

PR #977, opened 2015-03-13
Remove hardcoded version of the SQ Maven plugin

bellingard:
Hardcoding versions of the SQ Maven plugin prevents to analyse the project on future releases of SonarQube. For instance, the analysis of Titan on Nemo (http://nemo.sonarqube.org/dashboard/index?id=com.thinkaurelius.titan%3Atitan) is broken because of this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -322,11 +322,6 @@
                 </plugin>
                 <plugin>
                     <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>sonar-maven-plugin</artifactId>
-                    <version>2.1</version>
-                </plugin>
-                <plugin>
-                    <groupId>org.codehaus.mojo</groupId>
                     <artifactId>wagon-maven-plugin</artifactId>
                     <version>1.0-beta-4</version>
                 </plugin>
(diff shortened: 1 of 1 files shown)


bellingard:
FYI, with this modification, I managed to successfully analyze Titan on my latest SonarQube local installation.

dalaro:
Nemo compatibility is important.  Can we do this in a way that avoids this Maven warning about leaving the sonar-maven-plugin version unspecified?

```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for com.thinkaurelius.titan:titan:pom:0.5.5-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.codehaus.sonar:sonar-maven-plugin is missing. @ line 1026, column 29
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
```

bellingard:
@dalaro Indeed, removing the reference to the SQ Maven plugin should do the trick.

dalaro:
Thanks!  Worked great when I tried it out here, merged.

PR #978, opened 2015-03-15
Master

SongHuainan:
good


mbroecheler:
What is this?

dalaro:
This is still as incomprehensible to me now as it was back in March.

Issue #981, opened 2015-03-18
When change the TTL of Vertex,Edge,Propery,it can to zero


boliza:
When change the TTL of Vertex,Edge,Propery,it can set to zero.

dalaro:
Let's discuss this in the linked PR (#982) for now.  I just replied there.

dalaro:
Summarizing the discussion in #982: the point of this issue is to allow Titan to remove a finite TTL from an existing type, such that new elements of that type have no TTL.  I don't think #982 can be merged as-is, but the goal is still feasible.

I played with treating `setTTL(-1)` and `getTTL` returning `-1` as indicators of live-forever status on a type, but it was tricky to root out existing assumptions about 0 signifying live-forever, and to interop with Duration, which does not accept negative scalars.  So I stuck with 0 to signify live-forever.  This is implemented, along with three new tests, in the issue_981 branch.

dalaro:
I accidentally wrote #971 instead of #981 into multiple commits on the issue_981 branch.  I rebased it as issue_981_bis to fix those typos.  Matthias answered my only lingering question in that branch about an assertion change (https://github.com/thinkaurelius/titan/commit/addb1559b976c172d8297361c7e9a9ef3d940749#commitcomment-10968686).  I think this is ready to be turned into a PR and merged.

dalaro:
Merged #1069

Issue #92, opened 2012-09-20
Read robustness in Titan


mbroecheler:
Titan will automatically re-attempt saving transactional state upon encountering a TemporaryStorageException. It should do the same when reading.

mbroecheler:
fixed in "storage" branch

PR #982, opened 2015-03-18
Resolve issue #981

boliza:
Resolve issue #981

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/util/encoding/ConversionHelper.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/util/encoding/ConversionHelper.java
@@ -11,16 +11,17 @@ import java.util.concurrent.TimeUnit;
  */
 public class ConversionHelper {
 
-    public static final int getTTLSeconds(Duration duration) {
-        Preconditions.checkArgument(duration!=null && !duration.isZeroLength(),"Must provide non-zero TTL");
-        long ttlSeconds = Math.max(1,duration.getLength(TimeUnit.SECONDS));
-        assert ttlSeconds>0;
-        Preconditions.checkArgument(ttlSeconds<=Integer.MAX_VALUE, "tll value is too large [%s] - value overflow",duration);
-        return (int)ttlSeconds;
+    public static int getTTLSeconds(Duration duration) {
+        if (duration == null) {
+            return 0;
+        }
+        long ttlSeconds = Math.max(0, duration.getLength(TimeUnit.SECONDS));
+        Preconditions.checkArgument(ttlSeconds <= Integer.MAX_VALUE, "tll value is too large [%s] - value overflow", duration);
+        return (int) ttlSeconds;
     }
 
-    public static final int getTTLSeconds(long time, TimeUnit unit) {
-        return getTTLSeconds(new StandardDuration(time,unit));
+    public static int getTTLSeconds(long time, TimeUnit unit) {
+        return getTTLSeconds(new StandardDuration(time, unit));
     }
 
 }


dalaro:
Why is this necessary?  At first I assumed this was on the read path -- an edge case where an element comes back as it is expiring.  However, that doesn't seem to be the case.

This PR modifies one of two `ConversionHelper.getTTLSeconds` method overloads.  In practice it modifies both, since the unmodified overload just does some trivial parameter conversion and then calls the modified overload.

The two `getTTLSeconds` methods collectively have two call sites (not counting one overload calling the other):
- Setting the TTL on a `KCVSLogManager`'s store.  I don't see why we need 0 TTL here.  Isn't that the same as laboriously writing the log to `/dev/null`?
- Turning the `ttl` and `unit` parameters to `ManagementSystem.setTTL(type, ttl, unit)` into seconds.  This controls the TTL at the schema level.  My concern here is in the same vein as my concern for `KCVSLogManager`; what is the use of a type that dies as it is written?

Can you help me understand why this change is necessary?  Also, can you confirm that the #92 issue reference in the commit message was a typo?  I'm assuming you meant #982 or #981.

boliza:
> Can you help me understand why this change is necessary? Also, can you confirm that the #92 

this is my mistake ,sorry! It is means 981

boliza:
@dalaro You are right.It's not necessary, but sometimes when want store some data with ttl before forever.The only way we should do is ,add new property

dalaro:
> It's not necessary, but sometimes when want store some data with ttl before forever.The only way we should do is ,add new property

I'm struggling to understand.  Can you rephrase this?  I still don't understand why zero-ttl (instantly deleted) elements are useful.

boliza:
I am not good at English.When I design my graph , I think some properties should have ttl,but after a period of time ,because of the change of business（requirement）, those properties with ttl must be store without ttl.This just my requirement.You can think about it.

Thanks for you job!!!

dalaro:
Thanks, that's very helpful, I think I finally understand.

Titan's API is asymmetrical here:
- Javadoc for `TitanManagement.getTTL` says 0 TTL means live forever:
  
  ```
  Retrieves the time-to-live for the given {@link TitanSchemaType} as a {@link Duration}.
  If none has been explicitly defined, a zero-length {@link Duration} is returned.
  ```
- Javadoc for `TitanManagement.setTTL` is silent about what 0 TTL means.  The implementation treats 0 as invalid in practice and throws this exception.

This is a deficiency in Titan: `setTTL` and `getTTL` should interpret all values the same way, as you (reasonably) expected with 0.  I think I would actually prefer to make -1 mean "live forever" and disallow 0 (since 0 is more ambiguous than -1).

@boliza I have a separate concern about your use case.  Even if we make `TitanManagement.setTTL(<some-type>, <live-forever>)` possible, it may not be possible to update TTL on existing data.  Old data of `<some-type>` could expire even after changing the TTL.  New elements written after committing `TitanManagement.setTTL(<some-type>, <live-forever>)` should live forever.  Does that make sense?

boliza:
You are right,can't change the old data,just affect the new elements.

joshsh:
You are probably aware, but Titan inherits 0 as the default, infinite TTL from Cassandra. See [CQL v3](http://cassandra.apache.org/doc/cql3/CQL.html).  It might be wise to maintain that agreement, even to the extent of ensuring that negative values are also treated as infinite.  setTTL should indeed accept 0, particularly if TTL is mutable.

dalaro:
@joshsh Indeed, I abandoned the -1 idea once I got into the implementation.  It's not only at odds with Cassandra convention, but also a poor fit for Titan internals.  StandardDuration only accepts nonnegatives by design, and 0 is already hardcoded in several places as a "live forever" default value.  I recently replied about that on related issue #981.  Sorry to jump back and forth between here and there.

dalaro:
I think this has been superseded by #1069.  Please reopen #981 for comment if I missed something.

Issue #593, opened 2014-02-17
Support different cardinality in Indexing


mbroecheler:
Provide support for list and set properties by following:
- using the automatic array type of ES
- http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-update.html (search for array)
- be careful with the script syntax

Requires that the deleted value is made part of the IndexMutation as well.

Issue #736, opened 2014-08-28
Allow disabling and/or sandboxing the ES dynamic update script


dalaro:
The ES backend requires `script.disable_dynamic=false`.  ES recommends migrating to some combination of sandboxing and server-side, implicitly-trusted scripts: http://www.elasticsearch.org/blog/scripting-security/.  This is not something we absolutely have to do, but it would allow Titan to use ES in a broader range of client environments.

Titan appears to require dynamic scripting to support deletions of individual fields within a document, which are handled by StringBuilder'ing a script that calls `ctx._source.remove` on each target field, separating each remove call by semicolons.  If this is the only use site for dynamic scripting, then I don't think dynamic scripting is really a hard requirement.

We could probably add a config option that disables Titan's StringBuilder dynamic script logic and instead assumes the scriptfile is already installed on the server and refers to it by name, as in http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/modules-scripting.html.  Titan would ship with the script, but the user would have to manually install the script on their ES cluster before using this option.  On the upside, the user could then disable dynamic scripting (as far as Titan is concerned).

There are still some uncertain details: This would require the ability to pass a list or array of fields-to-delete as a script parameter binding over the wire.  It would also require that the server-side script iterate over the list/array passed as a bound script param.  And it's all got to work with `BulkRequestBuilder`.  This would require test coverage.

PR #994, opened 2015-04-03
Added support for different cardinality indexing in ES.

BrynCooke:
https://github.com/thinkaurelius/titan/issues/593
ES mutations are now done by script except if it is a new document.
Tests added.
Index.supports is called upon index creation to check if the key information supported. This results in a better error message.

@dalaro Please can you review the code changes.

### DIFF
--- a/docs/elasticsearch.txt
+++ b/docs/elasticsearch.txt
@@ -15,6 +15,7 @@ Titan supports http://elasticsearch.org[Elasticsearch] as an index backend.  Her
 * *Numeric Range*: Supports all numeric comparisons in `Compare`.
 * *Flexible Configuration*: Supports embedded or remote operation, custom transport and discovery, and open-ended settings customization.
 * *TTL*: Supports automatically expiring indexed elements.
+* *Collections*: Supports indexing SET and LIST cardinality properties.
 
 Please see <<version-compat>> for details on what versions of ES will work with Titan.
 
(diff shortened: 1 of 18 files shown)


BrynCooke:
@dalaro Hold on. The upgrade to ES1.5 has broken this.
Apparently script updates are now different.

BrynCooke:
I think this is to do with scripting being disabled:
http://www.elastic.co/guide/en/elasticsearch/reference/1.x/modules-scripting.html

dalaro:
Last time I looked at this (#736), I wasn't convinced that dynamic scripting was actually a hard requirement.  Perhaps we could rework field deletion to avoid it.

BrynCooke:
@dalaro I think this is good to go (After merging with the titan09 branch).

Dynamic scripting is already a requirement for Titan so to removing it should be pulled in to a different issue. It can probably be done, but would deployment of script files to every ES node in the cluster. I think we should wait for people to ask for this.

I think that the script language has either changed or the detection has changed from the earlier version of ES. It just so happened that the previous scripts were valid groovy and javascript.

Secondly there are race conditions in the tests. Particularly ElasticSearchConfigTest but anywhere where we embed ES could be affected.

As far as I can tell programmatically starting and stopping a node is not a synchronous operation, and examining the log files for started and stopped markers is insufficient to prevent conditions where:
1. An ES node is queried before it is ready.
2. An ES node joins a cluster that has been shut down.

This manifests itself in failed queries, data from previous tests replicating to new tests, nodes failing to start up due to the 'old master' node shutting down.

I have tried to make this better by pulling out the nodes in to separate clusters for each test and adding a delay after ES has started, but occasionally the tests still fail.

PR #999, opened 2015-04-10
Tweak the signatures of the internal titan interfaces

BrynCooke:
to prevent loss of context when constructing vertex and graph centric queries.

Otherwise multiple method calls on a query results in loss of typing.
Before:

```
Iterator<TitanVertex> it = tx.query().vertices();
Iterator<TitanVertex> it = tx.query().has("test").vertices();//Compile error
Iterator<Object> it = tx.query().has("test").vertices();// :0(
```

After:

```
Iterator<TitanVertex> it = tx.query().has("test").vertices();//OK
```

@mbroecheler @dalaro Consider merging this before making more M8 changes if you think the change is a good idea.

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/core/BaseVertexQuery.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/core/BaseVertexQuery.java
@@ -24,7 +24,7 @@ import org.apache.tinkerpop.gremlin.structure.Vertex;
  * @see TitanMultiVertexQuery
  * @author Matthias Br&ouml;cheler (http://www.matthiasb.com)
  */
-public interface BaseVertexQuery<Q extends BaseVertexQuery<Q>> {
+public interface BaseVertexQuery<Q> {
 
     /* ---------------------------------------------------------------
     * Query Specification
--- a/titan-core/src/main/java/com/thinkaurelius/titan/core/TitanGraphQuery.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/core/TitanGraphQuery.java
@@ -15,7 +15,7 @@ import org.apache.tinkerpop.gremlin.structure.Order;
  * @since 0.3.0
  */
 
-public interface TitanGraphQuery<Q extends TitanGraphQuery<Q>> {
+public interface TitanGraphQuery {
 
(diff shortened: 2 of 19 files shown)


dalaro:
@BrynCooke  Can you please elaborate?  I don't understand.  On titan09 HEAD at the moment, the middle line in your example doesn't compile since the RHS evaluates to Iterable:

```
Iterator<TitanVertex> it = tx.query().has("test").vertices();//Compile error
```

But this does compile (IDEA doesn't underline it and the Maven build compiles it without errors):

```
Iterator<TitanVertex> it = tx.query().has("test").vertices().iterator();
```

However, you seem to be getting at something more fundamental than this minor Iterator/Iterable difference, since the PR involves stripping the self-referential bounded type parameter on query classes.

BrynCooke:
@dalaro I am no longer sure about this:

Idea was flagging stuff like this for me

```
TitanVertex v = graph.addVertex();
        for (TitanVertex prop : v.query().has("name").vertices()) {

        }
```

As the self referencing was getting unwrapped on the first has method which then leaves you with an untyped generic as your context.

public TitanVertexQuery<? extends TitanVertexQuery> query();

The internal TitanVertexQuery is not generified a call to a method returns.

```
TitanVertexQuery
```

but actually needs

```
TitanVertexQuery<? extends TitanVertexQuery>
```

To achieve this using the existing method you would have to write:

```
public TitanVertexQuery<? extends TitanVertexQuery<? extends TitanVertexQuery> query();
```

Every method call unwraps until you end up with no type.

dalaro:
@BrynCooke I've been squinting at your reply, but I've failed to grasp the point.  I got completely lost at end with the suggestion about two layers of nested self-referential type parameter bounds.

Just to back up for one moment: does javac (run via the usual Maven commands) throw compile errors for the problem in question?  Or is it just IDEA's red underliner that complains?

BrynCooke:
Difficult to tell. It looks like we don't get all the errors in javac until
we correct some of the previous errors. Leave this for now. We can come
back to it later.

BrynCooke:
That is to say when we have fixed the previous errors I suspect we'll get a
bunch more that were not reported before.

BrynCooke:
I think this was an IDEA complaint.

Issue #987, opened 2015-03-24
Support Solr 5.0


BrynCooke:
Solr 5.x has removed various types that are used in schema.xml. 

Where X is Int, Float, Long, Double or Date

XField becomes TrieXField
SortedXField becomes TrieXField

I am not sure what the impact will be if any.

mbroecheler:
Is that something we should investigate for Titan 1.0 and upgrade to Solr 5.x? What's your recommendation here?

BrynCooke:
I don't think we can upgrade to Solr 5.0. Instead we can upgrade to 4.10.4 Which matches the version of Lucene that elastic search requires.

BrynCooke:
Actually we can upgrade to Solr 5.0.
We only use the Solr client rather than embedding it so it is fine to co-exist with older versions of lucene.

PR #874, opened 2014-12-10
update sample solr schema.xml

ghost:
Hi,

I made changes to the example and test schema.xml  files for solr to remove deprecation warnings when starting solr:

2096 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.IntField]. Please consult documentation how to replace it accordingly.
2101 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.LongField]. Please consult documentation how to replace it accordingly.
2106 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.FloatField]. Please consult documentation how to replace it accordingly.
2110 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.DoubleField]. Please consult documentation how to replace it accordingly.
2116 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.SortableIntField]. Please consult documentation how to replace it accordingly.
2120 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.SortableLongField]. Please consult documentation how to replace it accordingly.
2124 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.SortableFloatField]. Please consult documentation how to replace it accordingly.
2129 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.SortableDoubleField]. Please consult documentation how to replace it accordingly.
2138 [coreLoadExecutor-4-thread-1] WARN  org.apache.solr.core.SolrResourceLoader  – Solr loaded a deprecated plugin/analysis class [solr.DateField]. Please consult documentation how to replace it accordingly.

Tested this with Titan 0.5.2 and Solr 4.8.1:
gremlin> g=TitanFactory.open('./david.properties')
==>titangraph[cassandra:[9.30.182.235]]
[...]

PR #1002, opened 2015-04-11
Update solr configuration

randomstatistic:
With apologies for the wall of text, this does two things:
1. Addresses Issue #987 
2. Fixes the solr config files shipped with the Titan distribution. These had diverged from those in the test suite and were fundamentally broken.

I basically took the Solr 4.8.1 example config, copied over anything that looked necessary for Titan, and then cut out a bunch of the boilerplate/example stuff. Since I didn't know your policy on symlinks in git, I then just copied the modified config into the distribution.

Here's my notes on the changes from a Solr perspective:

schema.xml
- Replaced the deprecated Sortable<Type>Field types with Trie<Type>Field.
  I’m assuming the RangeQuery support is important, and used precisionStep=8 in the definitions to aid range query performance at some cost to index space.
- The Sortable<Type>Field definitions had sortMissingLast=“true”. I have not copied this, rather taking the default for the Trie fields, which "places docs without the field first in an ascending sort and last in a descending sort”. If Titan generally queries by "score desc", I think this should be safe. Please verify.
- Cut out most sections unrelated to what I thought Titan used out of the box, including all the multi-language definitions.
- Commented out FieldType definition sections that required extra files that weren’t really being used. (protwords.txt, etc) Users can uncomment these lines if they wish, and provide far more relevant files than those shipped in the Solr examples. 

solrconfig.xml
- Was there a reason you were using the ManagedIndexSchemaFactory to allow API-driven schema changes? Looks like you’re just failing over to schema.xml in the tests anyway, so I did not maintain this setting.
- Did you need unlockOnStartup=“true”? I’ve never needed this, and it felt unsafe. I did not maintain this setting.
- Removed several request handlers that were purely for example purposes.
- In the TTL support’s updateRequestProcessorChain definition, I’m uncertain why the fieldName is “expire_at_dt”, because I can’t see that name has ever existed. I kept it, it seemed harmless.
- Kept the ComplexPhraseQParserPlugin reference, since it felt like that explained the Solr >=4.8.1 requirement.

I have not tested this config with Solr 5.0, but I'd think it should work. Titan-Solr tests pass.


randomstatistic:
Ah, I just noticed PR #874. This PR logically contains that one, allows SolrCloud to function, and also brings the test and distribution copies of config in sync with each other and Solr 4.8.1.

randomstatistic:
It looks like you made some improvements to the Solr doc page since I originally went through the process, but I went ahead and added some of the stuff I've gleaned about the required Solr config for proper Titan support.

BrynCooke:
Looks good. Tested the config to create a new Solr core and connected with Titan and added some test data. Verified that the data made it in to Solr.

Issue #971, opened 2015-02-14
Fix encapsulation order


mbroecheler:
The nutshell explanation is that https://github.com/thinkaurelius/titan/commit/278d8b988058183ed032b03959771864f6b37a0c switched the encapsulation order of the Metrics and TTL StoreManager wrappers, but the Metrics wrapper didn't implement an interface expected by the TTL wrapper (the CustomizeStoreKCVSManager interface with the openDatabase(name, ttl) overload).  HBase still implements the interface expected by the TTL wrapper, but it's effectively hidden by the Metrics wrapper now sitting between them in the encapsulation order.  This leads to an exception on HBase with Metrics enabled as Backend tries to open a KCVSLogManager.

I tried to apply the simplest fix possible in https://github.com/thinkaurelius/titan/commit/d78a0801de473a1943061ef616d0252b55b98f66.  Mixing the decorator pattern and instanceof is likely to bite us again, and it might be worthwhile to try an alternate approach.  But I don't want to go nuts with refactoring right before a release, so I applied a low-effort fix.

dalaro:
I pushed a branch containing a potential fix for this (issue_971), but it adds nearly as many lines as it removes.  This may not even be worth doing.  I also moved the milestone to titan09, since the approach I took added an `openDatabase` overload signature to the KCVS interface.

mbroecheler:
If it doesn't simplify the code, then I'd say we drop it.

dalaro:
I reworked this and pushed new commits on the issue_971 branch.  I'm not going to merge it until all TTL tests are passing, and there's at least one preexisting failure that I need to resolve first (#1036).

Anyway, in the branch right now, KCVS defines two openDatabase methods:
- `openDatabase(String name, StoreMetaData.Container options)` with no default impl
- `openDatabase(String name)` with a default impl that calls the preceding method with an empty options container

The `StoreMetaData.Container` class uses the typesafe heterogeneous container pattern.  This means store impls don't have to cast TTL's associated object to an Integer anymore.  The only StoreMetaData key is TTL, though we could add new ones in the future.  Like EntryMetaData, StoreMetaData is not open to third-party extension, since core has to support any new members for them to be useful.

[...]

PR #1049, opened 2015-04-28
Rework TTL to avoid instanceof & casting

dalaro:
#971


dalaro:
I've reenabled some disabled TTL-related `@BrittleTests` in my local clone.  I want to see if I can find any interesting failures before merging this.  I'll update when I have those results.

dalaro:
I'm unable to elicit *GraphTest TTL test failures on Cassandra or HBase even with all of the `@BrittleTests` annotations in TitanGraphTest commented.  Merging this in.

PR #1069, opened 2015-05-01
Allow removing a TTL (#981)

dalaro:
This PR changes `setTTL(type, 0, unit)` to mean: remove any existing TTL on the named type, such that new elements of said type will not automatically expire.  Existing elements of the type may still die by TTL.

The scariest line change in here is probably the assertion deletion in `AbstractVertex.remove`.  That was cleared by Matthias: https://github.com/thinkaurelius/titan/commit/addb1559b976c172d8297361c7e9a9ef3d940749#commitcomment-10968686.


boliza:
:+1:

PR #1076, opened 2015-05-05
Removed custom temporal classes in favour of the standard Java8 temporal classes

BrynCooke:
- Removed custom temporal classes in favour of the standard Java8 classes
- All serialisation to epoch units is now left to the timestamp provider rather than temporal classes carrying around their units.
- Indexes now support Instant in addition to the legacy Date class.
- TTL now uses Duration in managment to give the user a symmetic API.

This greatly simplifies all of the temporal code in the codebase
I didn't remove Date support form indexes as this would be a bit mean to folks that exclusively use legacy APIs.

@mbroecheler @dalaro 
This is a major change, before merging I need:
- Approval from you guys
- To rerun the unit tests. There were quite a few merge conflicts when I rebased so I need to ensure that everything is still working.
- Confirmation that if an index commit fails then the errors are output to the logs but the rest of the transaction proceeds is the desired behaviour.
- To get the HBase tests running. I have never got these running locally.


mbroecheler:
to be reviewed by @dalaro

dalaro:
Can you elaborate on the concern around failed index transactions?  I'd like to understand why this is singled out as a potential problem area.

I took a couple of passes through the diff.  I committed two changes atop your current PR HEAD.
1. Tweaking `DistributedStoreManager.MaskedTimestamp.getAdditionTime`.  This has a couple of overloads.  The old behavior of both overloads would convert the timestamp to a long and set the least significant bit.  One of the two overloads of this method retained this behavior; the other, seemingly, did not.  This has no effect on microsecond and nanosecond timestamps, but can have an effect (i.e. failed tests and mysteriously missing writes) with millisecond timestamps.  While tweaking it, I made the signatures ambiguous, so I changed one overload to `getAdditionTimeInstant`.  https://github.com/dalaro/titan/commit/136f54cd991ab8126e05e6627a384a2ce519c165
2. Deleting a vestigial comment. https://github.com/dalaro/titan/commit/04ee9c714f4f86c8e3f0b906a7b06d4883bdd82f.

Both of these commits are in git://github.com/dalaro/titan.git TITAN_1019.  They're based on your https://github.com/BrynCooke/titan/commit/a190c607ad1b1981bd1337c0f1442aab9ee05a08, so they should fast forward merge if you pull from my remote.

The BDB tests completed, and while I ran into some Cassandra failures, I think they're all preexisting on titan09.  I haven't run with HBase.

BrynCooke:
@dalaro Regarding the index transactions, if there is a permanent exception thrown from within the mutate method, the transaction will still proceed. The reason I came up against this is that Solr doesn't support nanosecond granularity indexing. So I throw an illegal argument exception if this is detected in the mutate method. However, this exception caught and output to the logs leaving the possibility that the index will go out of sync with the underlying data.

BrynCooke:
I've merged @dalaro 's changes and squashed them in to the original commit. The index transactions issue is a separate issue where behaviour could be improved by allowing an index to verify that a value is supported at the point where the user assigns a property.

PR #1092, opened 2015-05-21
Renamed the Json to GraphSON.

dmill-bz:
The Json serializers have been renamed in M9, changed the config file to reflect this. The new serializers go by GraphSON.
gremlin-server will throw warnings without this change. With that said, I think GraphSON is now default and doesn't have to be explicitly set in the configuration. It might be safe to remove these two files.


dmill-bz:
I also changed the name of the graph from `g` > `graph`
`g` is pretty confusing as it is generally used to traverse aka: `graph.traversal()`. I changed the name in the configuration file to `graph` so that it would follow in the steps of the gremlin-server configuration and is more intuitive for the users. 
Anyone who needs to traverse `g` can set a script to do so. (maybe we should set one by default? `g = graph.traversal()` ?)

I'm not sure if this breaks anything so let me know.

spmallette:
Thanks for this.  I think we should include an init script to give us `g`.  i'll get that added.

spmallette:
I just pushed the change to titan09 branch for binding "g" to the script engine.  please feel free to try it out.

dmill-bz:
Work as expected! Thanks !!

PR #1093, opened 2015-05-21
static predicates for Text and Geo

pluradj:
TP3 changed the syntax of the has() step between M8 and M9. In M8, there was `has(key,bi-predicate,object)` but in M9, there is only `has(key,predicate)`.

So right now with the latest TP3 master and titan09, you have to do something like this:

``` java
g.V().has('name', new P(CONTAINS, 'jason'))
```

With this PR, you can do a has() query with a bi-predicate like this

``` java
g.V().has('name', textContains('jason'))
```

This PR is based on the statics found in here -- https://github.com/apache/incubator-tinkerpop/blob/master/gremlin-core/src/main/java/org/apache/tinkerpop/gremlin/process/traversal/P.java

I'm not clear on what should happen with Cmp.java or Contain.java since they seem to have duplicate function of what P.java already has.

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/core/attribute/Geo.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/core/attribute/Geo.java
@@ -2,6 +2,7 @@ package com.thinkaurelius.titan.core.attribute;
 
 import com.google.common.base.Preconditions;
 import com.thinkaurelius.titan.graphdb.query.TitanPredicate;
+import org.apache.tinkerpop.gremlin.process.traversal.P;
 
 /**
  * Comparison relations for geographic shapes.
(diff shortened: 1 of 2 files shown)


dalaro:
LGTM.  Sorry for the very long reply delay.  Can you please accept the Titan CLA (http://titan-cla.datastax.com/) before I merge?

pluradj:
Done! Thanks, Dan. Glad to see you back.

dkuppitz:
> I'm not clear on what should happen with Cmp.java or Contain.java since they seem to have duplicate function of what P.java already has.

Good point. @dalaro? Will Titan still use the index if we use the TP3 standard predicates like `P.eq` instead of `Cmp.EQUAL` or `P.within` instead of `Contain.WITHIN`?

dkuppitz:
I just tried it out. At least I don't get an index warning. So it looks like we can get rid of `Cmp` and `Contain` altogether.

mbroecheler:
Cmp and Contain are used internally by Titan. However, Titan supports all the TP3 predicates and decomposes them into Cmp and Contain based on the query optimizer.

So, you can get "rid" of them in your query code but they will remain in the codebase.

PR #1101, opened 2015-06-08
specify version of hppc

mpollmeier:
In a project that depends on titan-core the version of hppc was automatically resolved to 0.7.1 (the latest), which isn't compatible. Fixing it to 0.6.0 should work.

### DIFF
--- a/titan-core/pom.xml
+++ b/titan-core/pom.xml
@@ -94,6 +94,7 @@
         <dependency>
             <groupId>com.carrotsearch</groupId>
             <artifactId>hppc</artifactId>
+            <version>0.6.0</version>
         </dependency>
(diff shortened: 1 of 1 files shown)


mpollmeier:
I've used 0.6.0 because that's what mvn dependency:tree resolved to in titan-core. 0.6.1 works aswell, so maybe we should use that.

dalaro:
titan09 now uses 0.7.1 (as of 9a49faa3669fc2c03e458c7999ecf095e0b0725d).  The dependency version is currently controlled in the root pom.xml's `<dependencyManagement>`.

mpollmeier:
Sweet, thank you!

PR #1120, opened 2015-07-12
Modify SEPARATOR_CHAR(% -> 0x1e) and SPECIAL_TYPE_CHAR($ -> 0x1f)

titsuki:
The RDF data such as dbpedia(en) sometimes contains URI that uses dollar symbols or percentage symbols.
- example 1

```
<http://dbpedia.org/property/gdpPerCapitaUs$_>
```
- example 2

```
<http://dbpedia.org/property/gemeindeschl%C3%BCssel>
```

Those symbols cause "Name can not contains reserved character %:" or "Name can not contains reserved character $:" Errors when you import the RDF data to the database.
Therefore, I modified SEPARATOR_CHAR % to 0x1e(RecordSeparator), and SPECIAL_TYPE_CHAR $ to 0x1f(UnitSeparator).

I think those are more suitable for reserved characters.

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/internal/Token.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/internal/Token.java
@@ -8,9 +8,9 @@ import org.apache.commons.lang.StringUtils;
  */
 public class Token {
 
-    public static final char SEPARATOR_CHAR = '%';
+    public static final char SEPARATOR_CHAR = 0x1e;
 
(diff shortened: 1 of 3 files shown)


mbroecheler:
Thanks for the feedback - this is useful. We have since moved away from '$' to use TP3 hidden namespace but I have adopted your suggestion for using the record separate as a token separator. So, hopefully this should be fixed for you in Titan 1.0. Definitely test it when we release the snapshot next week.

mbroecheler:
Commit: c9272c86c2f34ec09d21a883ca91eb019a6392a4

PR #1121, opened 2015-07-13
Fix bulk operation error propagation for elasticsearch

timwu:
In the event of a failure in a batch of operations, elastic search
appears to return a bulk result (for the failing operation) with
a failure object. However when there is no failure, the failure
object is null. The current handling code for bubbling up possible
exceptions does not account for the case where there are successful
operations in a batch of operations containing a failure, leading
to a NPE. This patch adds a test to demonstrate the issue and adds
a check for failure before examining the bulk result.


nichochar:
+1

dalaro:
LGTM.  I appreciate both the description of how the NPE can fall out of a multi-op batch with partial failure and the test cases.  @timwu: Can you please accept the Titan CLA (http://titan-cla.datastax.com/) before I merge?

Issue #1110, opened 2015-06-22
Upgrade hppc version to match gremlin-core


dalaro:
Titan and gremlin-core depend on different versions of HPPC.  Titan compiles with 0.6.0.  gremlin-core compiles with 0.7.1.  Incompatible API changes in HPPC 0.7.0 make Titan unable to run with gremlin-core's preferred HPPC version: https://groups.google.com/d/msg/aureliusgraphs/ycd8ddkurYw/iSrNiaqsb04J.

The missing class in that thread didn't go away.  It was just renamed to LongHashSet.  I suspect the HPPC API changes don't affect Titan beyond some class renames (but I haven't investigated).  We should consider putting Titan back in HPPC version lockstep with gremlin-core and/or shading Titan's HPPC version.  Titan's version of HPPC is controlled in the top-level pom.xml.

PR #1122, opened 2015-07-14
update hppc version to match tp3 version

fitzoh:
#1110

Simple find and replace of hppc class names.

Passed titan-core, couldn't get titan-cassandra working (configuration issues).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -916,7 +916,7 @@
             <dependency>
                 <groupId>com.carrotsearch</groupId>
                 <artifactId>hppc</artifactId>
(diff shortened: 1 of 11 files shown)


dalaro:
This was conflicted when I picked up #1110.  It was faster to just find-replace again than to try to resolve conflicts, which I committed as 9a49faa3669fc2c03e458c7999ecf095e0b0725d.  I'm unclear on the titan-cassandra issues... the test suite seems to be unaffected by this change so far, but I'm only a few classes into the full -cassandra suite at the moment.

PR #1124, opened 2015-07-22
Update Geoshape.java

bravecharge:
Add to graphson format deserialization support by regular,
there will be problems when importing grapson file Before,


mbroecheler:
Can you elaborate on why this is needed and provide a unit test to support it? That would be excellent - thank you!

Issue #1125, opened 2015-07-23
Thift framesize out of bound


mishraawake:
We are facing one problem herein for some shallow node, default thift frame size exceeds from its limit defined in asyntax client. 
Is titan planning to include latest as astyanax library because in new  astyanax  library there is provision of configuring this particular setting that we need to increase.

Version of titan that we are using is : 0.5.4

There is another work around for this problem is to use cassandrathrift as backend configuration.
But I don't think it is recommended for live environment.

For your kind information we are getting following exception.

com.thinkaurelius.titan.core.TitanException: Could not execute operation due to backend exception

```
   at com.thinkaurelius.titan.diskstorage.util.BackendOperation.execute(BackendOperation.java:44)

   at com.thinkaurelius.titan.diskstorage.BackendTransaction.executeRead(BackendTransaction.java:428)

   at com.thinkaurelius.titan.diskstorage.BackendTransaction.edgeStoreMultiQuery(BackendTransaction.java:269)

   at com.thinkaurelius.titan.graphdb.database.StandardTitanGraph.edgeMultiQuery(StandardTitanGraph.java:375)

   at com.thinkaurelius.titan.graphdb.transaction.StandardTitanTx.executeMultiQuery(StandardTitanTx.java:1002)

   at com.thinkaurelius.titan.graphdb.query.vertex.MultiVertexCentricQueryBuilder.execute(MultiVertexCentricQueryBuilder.java:92)

   at com.thinkaurelius.titan.graphdb.query.vertex.MultiVertexCentricQueryBuilder.titanEdges(MultiVertexCentricQueryBuilder.java:112)

   at com.til.cms.graph.titan.read.dao.GraphReadDaoImpl.getChildHierarchyEdgesBasedOnRank(GraphReadDaoImpl.java:205)

   at com.til.cms.graph.titan.services.TitanGraphRead.getContentListNodes(TitanGraphRead.java:964)

   at com.til.cms.graph.titan.services.TitanGraphRead.getContentListByRank(TitanGraphRead.java:313)

   at com.til.cms.graph.controller.JCMSController.getSpecialSections(JCMSController.java:233)

   at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

   at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)

   at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

[...]

PR #1127, opened 2015-07-25
Update hppc to 0.7.1 to match TP3's dependency on hppc

jkschneider:
HPPC introduced a breaking change to Titan with this commit: https://github.com/carrotsearch/hppc/commit/3fe735b781b6b3f7cf87b205a1dea186b96c8c2e

Latest version conflict resolution in Gradle and Ivy cause Titan 0.9 M2 and Tinkerpop 3.0.0-incubating to be incompatible because of TP3's dependency on HPPC 0.7.1

Also contained in this PR is a change in Tinkerpop's version from 3.0.0-SNAPSHOT to 3.0.0-incubating now that it has been released.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
     </scm>
     <properties>
         <titan.compatible.versions />
-        <tinkerpop.version>3.0.0-SNAPSHOT</tinkerpop.version>
+        <tinkerpop.version>3.0.0-incubating</tinkerpop.version>
         <junit.version>4.12</junit.version>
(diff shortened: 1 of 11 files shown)


dalaro:
titan09 got the 3.0.0-incubating change from some other commit a while back.  The rest of this PR looks like #1125.   Both are conflicted by now.  It ended up being fastest to rerun the find-replaces to remove "Open" than to try resolving the conflicts in this particular case (9a49faa3669fc2c03e458c7999ecf095e0b0725d).

Issue #1126, opened 2015-07-23
Improved Solr Support


ghost:
The SolrIndex.java code responsible for providing the SolrCloud interface to Titan currently requires a config set name to match a collection name in SolrCloud.  This requires manual creation of many config sets in ZooKeeper before Titan mixed index commands can be issued.

A new config is not be required for every SolrCloud Collection created by Titan.  The same config set can be reused over and over again if the schema, for example, does not need to be different, which is often the case.

In addition, the SYNC state that is checked for in the  waitForRecoveriesToFinish method of the class was a) never really used and b) is no longer supported in Solr 5.2.1 and causes compile errors.
The SYNC code should be removed, allowing Solr 5.2.1 to be supported (currently the latest version of Solr).
http://mail-archives.apache.org/mod_mbox/lucene-solr-user/201410.mbox/%3CCAEvwVFEvCA=5=g7i2HpEtL=51HedoZksKYN+jYierPwWXpZkuA@mail.gmail.com%3E

A new property called 
       index.search.solr.configset
should be supported in the Titan properties file that allows a user to specify the name of a default SolrCloud config set that is reused repeatedly each time a new mixed index (Collection) is created.  This change will allow a user to upload a single config set manually into SolrCloud, and from then on, every time a mixed index statement is encountered, a new Collection is automatically created in Solr with no more manual interventions required.  The advantage of using a property, is that if the property is missing, the behavior of the Solr interface will be as it is today, allowing a different config set per Collection if that is actually needed.

A pull request will be submitted for this Issue shortly that addresses the new property as well as Solr 5.2.1 support and the removal of SYNC.

dalaro:
The linked PR was recently merged.  I think the merge concludes this issue, so I'm closing it.  Please reopen if I missed something.

PR #1132, opened 2015-08-17
TITAN 1126: upgrade to latest solr and improve solrcloud support

ghost:
https://github.com/thinkaurelius/titan/issues/1126

Adds a new property and additional support for SolrCloud to allow re-use of config sets and reduce amount of manual configuration needed to use Titan with Solr.

### DIFF
--- a/CHANGELOG.asc
+++ b/CHANGELOG.asc
@@ -20,7 +20,7 @@ Version 0.9.0-M2 (June 9, 2015)
 * BerkeleyJE 5.0.73
 * Elasticsearch 1.5.1
 * Lucene 4.10.4
-* Solr 5.1.0
+* Solr 5.2.1
 * Java 1.8
 
 *Features:*
--- a/docs/solr.txt
+++ b/docs/solr.txt
@@ -22,7 +22,8 @@ Solr Configuration Overview
 Connecting to Solr
 ^^^^^^^^^^^^^^^^^^
 
-Titan supports two different modes of connecting to a running Solr instance or cluster of instances: SolrCloud and HTTP. The desired connection mode is configured via the parameter `mode` which must be set to either `http` or `cloud`, the latter being the default value.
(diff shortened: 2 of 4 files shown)


dalaro:
This looks interesting.  I like the documentation changes.  If I'm reading the source changes right, this affects creation of new Solr-backed indices.  It introduces a new config option called `configset`.  If set, Titan will assume it names a extant directory containing config files in the Solr cluster, and Titan will setup collections it creates to reuse those config files.  Critically, it does not touch the actual name of the collection, just the name of a collection's configs.  So we can still have multiple indices per backend, but now the user has the option of writing and uploading Solr config files once and then reusing them across multiple collections.

Did I understand the PR?  If so, do you have any suggestions on how we should test this?

dalaro:
Also: can you please sign the new Titan CLA (http://titan-cla.datastax.com/)?

ghost:
You summarized the pull request/changes precisely.

One detail: Once the initial config set is loaded from disk into Zookeeper
by SolrCloud, future references to the config set are reused from Zookeeper
and not from disk.

ghost:
Dan,

You asked about verification.

I am thinking out loud with you here rather than proposing anything
specific at the moment.

The way I verify this feature manually is to create several mixed indexes
in a test Titan
database, and make sure the indexes are working as expected via the Gremlin
shell.

I then log into the Zookeeper cli - zkCli.sh and verify that the Solr Cloud
data there is as I would expect to see it.

Under /configs I expect to see only a single configset defined - the one
referenced in the Titan configuration file and the one copied under solr
cloud.  There should be no other config sets present :    ls /configs

Contrast this to the default behavior where multiple configsets will appear
in the ls /configs listing.

I manually check the collections in Zookeeper:  ls /collections to make
sure there is an entry for all of the indexes that were created in Titan.

I manually verify that there is only a single directory under the Solr
Cloud directory structure that contains the single Titan default config set
(or whatever config set I am re-using).  Contrast this to the default
behavior where multiple manual copies and subdirectories for each config
set are required.

That's about it.  What do you think ?

dalaro:
Thanks for replying about testing.  Accessing ZK is an interesting angle.  I wonder if there is an programmatically-accessible API that easily dumps enough config information for test assertions (maybe that's ZK in this case).  That's sort of how we test some config-manipulating options in ES.

This is a substantial quality of life improvement without, as far as I can tell, much of a potential surface area for bugs; and the fallback to pre-merge behavior in the default configuration is another plus.  I merged it into titan09.

Issue #1135, opened 2015-09-01
make optimisticLocking a store feature?


amcp:
These tests override isLockingOptimistic:
https://github.com/thinkaurelius/titan/blob/47e9b35efaa4d1eb8722c1db0b1e599336a2c049/titan-cassandra/src/test/java/com/thinkaurelius/titan/graphdb/CassandraGraphTest.java#L28
https://github.com/thinkaurelius/titan/blob/47e9b35efaa4d1eb8722c1db0b1e599336a2c049/titan-berkeleyje/src/test/java/com/thinkaurelius/titan/graphdb/berkeleyje/BerkeleyGraphTest.java#L62
https://github.com/thinkaurelius/titan/blob/47e9b35efaa4d1eb8722c1db0b1e599336a2c049/titan-test/src/test/java/com/thinkaurelius/titan/graphdb/inmemory/InMemoryGraphTest.java#L71
https://github.com/thinkaurelius/titan/blob/47e9b35efaa4d1eb8722c1db0b1e599336a2c049/titan-hbase-parent/titan-hbase-core/src/test/java/com/thinkaurelius/titan/graphdb/hbase/HBaseGraphTest.java#L32
https://github.com/thinkaurelius/titan/blob/47e9b35efaa4d1eb8722c1db0b1e599336a2c049/titan-cassandra/src/test/java/com/thinkaurelius/titan/graphdb/thrift/ThriftGraphCacheTest.java#L25

This value varies from one storage backend to another. Those differences are summarized [here](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Tools.TitanDB.html#Tools.TitanDB.TitanFeatures), and I think it makes sense to pull this out of tests into StoreFeatures. If you agree I could make a pull request...

Thanks
Alex

amcp:
created pull request https://github.com/amcp/titan/commit/0638987c7b5907a790e880e1eec305b49ac6494c

dalaro:
`isLockingOptimistic` was a hack in the test code before the PR, but if we're pulling it into src/main like this, then perhaps we should refine the concept a bit.  Thinking out loud here...

It might be misleadingly named, since "pessimistic" behavior on BDB only really applies to its read locks; write lock acquisition is still deferred until the titan transaction commits.

[...]

PR #1136, opened 2015-09-01
pull optimistic locking up into store features

amcp:
also cleaned up a bunch of unused import


dalaro:
Let's talk about it in #1135.

Also, please don't combine import cleanups with other changes.  It's all one commit, so there was no easy way for me to see just a diff of your optimistic locking feature tweak -- I had to scroll through a wall of banal import changes to find it.  Import cleanups are okay, but they're easier to handle in a separate commit (and preferably in its own PR).

PR #1142, opened 2015-09-04
Added missing break in ElasticSearchIndex, that now not all fields are stored analysed and not_analysed

deiflaender:
Added missing break in ElasticSearchIndex, that now not all fields are stored analysed and not_analysed
This is also contained in version 0.9


dalaro:
LGTM.  I agree that `TEXT` should use the default `index=analyzed` value; the docs on `Mapping` explicitly state that `TEXT` is intended to be tokenized.  I'll check to make sure this doesn't break any tests (it shouldn't).  Can you please accept the Titan CLA (http://titan-cla.datastax.com/) before I merge?

You're also right that this bug exists in both 05 and 09.

deiflaender:
Hi,

I signed the document...

regards

D. Eifländer

dalaro:
cherry-picked to titan09 as e2da8e3249d63261eefaabdb2c43e7a870252d0b (behind/conflicted on merging to the extent that it may not be viable).

PR #1146, opened 2015-09-07
Enabled to set the thift frame size for astyanax.

deiflaender:


dalaro:
Thanks for this.  I'm not sure whether to apply it on titan05, but I did just commit something similar to titan09: 1001f3a6b3c0a19222613d2385b212e7d618c89a.  I unified the option for frame size between the cassandrathrift and astyanax backends, and I went up to the latest current astyanax version 3.8.0.  Is there a reason the PR specifically chose 3.6.0 instead?

deiflaender:
Hi, it would be nice to have this in titan0.5, because
cassandrathrift-backend(obsolete now?) does not reconnect to cassandra
cluster if connections get lost.
The reason why i took 3.6.0, was only that 3.8.0 was not available as i
implemented this on our local branch(has some more minor fixes, pull
requests comming soon...)

dalaro:
I suppose we can defend the version upgrade on the grounds that it enables functionality crucial for accessing some datasets with Astyanax.  I'll look at applying this to titan05 as well.  For backwards compatibility on titan05, I will retain your option layout (a pair of duplicate frame-size options, one for astyanax and one for thrift).

concretevitamin:
@deiflaender just curious, are those "minor fixes" in your local branch related to Thrift frame size / Titan-Cassandra connection issues?

deiflaender:
No

PR #1153, opened 2015-09-15
Upgrade to ES2.0.0-beta1 and Lucene 5.2.1

jamesecahill:
I'm not sure of the state of the titan09 branch, but I couldn't get tests to run with or without this change set. The changes to 2.x are relatively simple, with the merging of queries and filters, and some other class re-org stuff. Please let me know if you think this is something you can include, or build on, for you 1.0 release.


zazzcode:
So does this mean the next release of Titan will support Elasticsearch 2.   Or do we need to build from source to include Elasticsearch 2

seeIT-52:
Sorry, I wonder can this PR merge into the 1.0 master? And when?

PR #1156, opened 2015-09-23
add CompactionStrategy options for Cassandra storage configuration

pluradj:
Allows org.apache.cassandra.db.compaction.LeveledCompactionStrategy to be chosen when the Titan tables are created.


pluradj:
https://groups.google.com/d/msg/aureliusgraphs/S6NvSUFpoxA/i0S7DQgQxVwJ

dalaro:
I've merged this into titan11.  After merging, I tweaked its default behavior in d647a6925658a3709899b521fca3fb23e082e1c8.  I was concerned about how the PR hardcoded SizeTiered as the default compaction strategy in the absence of a user-specified override.  My commit makes Titan behaves as it did before this PR, remaining silent on the strategy in the absence of a user-specified option, implicitly defaulting to whatever C\* or the client-side driver wants to use.

pluradj:
Makes sense. Thanks Dan.

PR #1157, opened 2015-09-23
Throw an exception when an Elasticsearch bulk request fails

srosenthal:
My company has some Titan instances that get a lot of traffic. When Elasticsearch requests time out or fail, Titan 0.5.x would ignore the failure. We'd prefer it to fail more loudly.

The latest 0.9.x versions have done something very similar here:
https://github.com/thinkaurelius/titan/blob/0.9.0-M2/titan-es/src/main/java/com/thinkaurelius/titan/diskstorage/es/ElasticSearchIndex.java#L600


ghost:
@dalaro:  I believe the premise of this pull request to be valid for Titan 1.1 - although it looks like 75% of this pull request is already implemented in the Titan 1.1 ElasticSearchIndex.java file.

The differences between the pull request and Titan 1.1 version of ElasticSearchIndex.java are that the logic in checkBulkResponseForFailures is "inlined" for the if (bulkrequests > 0)  case, and non-existent yet for the if (requests > 0) case.

My recommendation for Titan 1.1 is to create a pull request that essentially implements the code submitted here in the pull request - but with slight changes that acknowledge the imports, etc. are already in Titan 1.1 code.   In otherwords, just update the code here to fit Titan 1.1 current code and include it.  Can you please advise ?

Issue #1160, opened 2015-09-29
Document reserved system names


mbroecheler:
vertex
element
edge
property
label
key

spmallette:
The docs were updated in the the PR #1162 - closing.

PR #1162, opened 2015-09-29
adding reserved keywords to docs

elubow:
This deals with thinkaurelius/titan#1160 by adding a reserved keywords section to the docs.  Happy to cleanup further if you don't like the language.


dalaro:
Could you please make the following changes?
- Emphasize that the list of names is reserved for all schemata: vertex labels, edge labels, and property keys
- typo: regareding
- To sync with the rest of the docs, leave "Vertex" not capitalized unless it starts a sentence -- "vertex labels" is fine in the middle of a sentence

elubow:
Good catch on the typo. Added the rest of the schemata.  Also moved the example to make it more clear by using the word `property` with a `vertex`.  Squashed the change in to the original commit for clarity.

Issue #1169, opened 2015-10-12
NPE when using repeat step with query.batch=true


twilmes:
Version: Titan 1.0.0
This query works without issue with query.batch off.

Steps to reproduce using toy data with the relevant portion of the exception. 

``` groovy
gremlin> graph = TitanFactory.open('./conf/titan-cassandra.properties')
==>standardtitangraph[cassandrathrift:[127.0.0.1]]
gremlin> g = graph.traversal()
==>graphtraversalsource[standardtitangraph[cassandrathrift:[127.0.0.1]], standard]
gremlin> v1 = graph.addVertex('id', 1)
==>v[4192]
gremlin> v2 = graph.addVertex('id', 2)
==>v[4096]
gremlin> v1.addEdge('knows', v2)
==>e[170-38g-28lx-35s][4192-knows->4096]
gremlin> g.tx().commit()
gremlin> g.V().has('id', 1).emit().repeat(out('knows'))
10:10:49 WARN  com.thinkaurelius.titan.graphdb.transaction.StandardTitanTx  - Query requires iterating over all vertices [(id = 1)]. For better performance, use indexes
==>v[4192]
java.lang.NullPointerException
Display stack trace? [yN] y
java.lang.NullPointerException
    at com.thinkaurelius.titan.graphdb.tinkerpop.optimize.TitanVertexStep.flatMap(TitanVertexStep.java:95)
    at org.apache.tinkerpop.gremlin.process.traversal.step.map.FlatMapStep.processNextStart(FlatMapStep.java:47)
    at com.thinkaurelius.titan.graphdb.tinkerpop.optimize.TitanVertexStep.processNextStart(TitanVertexStep.java:88)
    at org.apache.tinkerpop.gremlin.process.traversal.step.util.AbstractStep.hasNext(AbstractStep.java:140)
    at org.apache.tinkerpop.gremlin.process.traversal.step.util.ExpandableStepIterator.next(ExpandableStepIterator.java:54)
    at org.apache.tinkerpop.gremlin.process.traversal.step.branch.RepeatStep$RepeatEndStep.standardAlgorithm(RepeatStep.java:230)
    at org.apache.tinkerpop.gremlin.process.traversal.step.util.ComputerAwareStep.processNextStart(ComputerAwareStep.java:47)
    at org.apache.tinkerpop.gremlin.process.traversal.step.util.AbstractStep.hasNext(AbstractStep.java:140)
[...]

PR #1170, opened 2015-10-13
NPE when using repeat step with query.batch=true

twilmes:
This PR is for issue #1169. I'm pretty new to the interaction between TP3 & the underlying Titan specific steps but I figured I'd take a crack at it. 

After debugging, it appeared that at time of the TitanVertex step's initialization (where the multi query would be traditionally executed if batch==true), the starter vertices were not yet available because of the vertex steps placement nested within the repeat.  I thought about attempting to work my way backwards through the traversal but that seemed problematic.  I instead opted for an additional check in the flatMap to see if the multi query results were really loaded during initialization.  If not, initialization is called again.

One additional thing to note, when putting the test together, it appeared that the TitanVertex step nested within the repeat did not have the expected profiling annotations.  I may have made the incorrect assumption, but thought that this may be a separate, unrelated issue so I commented that test out and added a note.  If this is incorrect, please let me know, or I can also enter another issue if there is something else awry with the profile metrics on nested steps.


twilmes:
It appears this does not fix all cases.  I tested with the emit step after the repeat and no exception is thrown, but no results are produced.

`gts.V().has('id', 1).repeat(out('knows')).emit()`

I will look into this and update the PR.

twilmes:
Closing this PR because more analysis needs to be done.

PR #1195, opened 2015-11-25
Handle self-loop edges in CassandraInputFormat.

adeandrade:
Otherwise an exception is raised because we are adding the same edge twice.


dalaro:
Thanks!

Prior to merging this into titan11, I added a test to cover this case (7244ededaec63a712191b9b385247b2035ed8865).  The test duly failed.  Merging this PR fixed it.

PR #1197, opened 2015-11-26
Make TitanVertexDeserializer (refCounter) a static object.

adeandrade:
A TitanVertexDeserializer instance  is being created by each InputFormat instance. TitanVertexDeserializer requires a StandardTitanGraph for property and schema inference. This connection is read-only and does not transfer significant data.

The number of these connections is growing linearly with the number of splits (partitions). In an usual scenario, this can be 2 orders of magnitude. Moreover, a Backend instance will retrieve its stored configuration and automatically establish unneeded index connections.

As an example, using Spark's NewHadoopRDD in a small Cassandra cluster, CassandraInputFormat tries to establish 513 cassandra and elasticsearch connections before it breaks.

Making TitanVertexDeserializer a static variable ensures only one instance will be created per process.


dalaro:
This is merged into titan11.

I like this optimization, but I had to tweak it a bit to make it play nicely with our tests.  Some of the tests run multiple jobs (via Hadoop's LocalJobRunner) with distinct configs without restarting the VM or resetting the static state added by the PR, leading to a spurious test failure caused by stale config data.  I modified GiraphInputFormat so that whenever the refcount rises from zero to one, it applies the configuration most recently passed to `GiraphInputFormat.setConf` anywhere in the VM (ead4a3d49d966a4d2ebad845a1c6c7b67ac709f2).

PR #1200, opened 2015-12-02
Serializer Documentation out of date

pkohan:
The documentation for serializers is out of date, Titan 1.0 appears to not support automatic serialization of HashMap or ArrayList. 

More than happy to change the phrasing here if you want, just wanted to get the misinformation out of there.


spmallette:
thanks @pkohan but it looks like there is a more complete PR out there for this already with #1194.  i think we will merge that one. i'm going to close this PR.

PR #1208, opened 2015-12-11
version update slf4j

ghost:
update version of slf4j in Titan to match TP3.1 version

built.  ran test cases.


dalaro:
Merged into titan11, thanks.

PR #1211, opened 2015-12-12
Update jersey

ghost:
update version of jersey

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
         <asm3.version>3.1</asm3.version>
         <asm4.version>4.0</asm4.version>
         <zookeeper.version>3.4.6</zookeeper.version>
-        <jersey.version>1.9</jersey.version>
+        <jersey.version>1.19</jersey.version>
         <jna.version>4.0.0</jna.version>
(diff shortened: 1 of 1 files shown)


ghost:
Looks like the current version of jersey (1.9) is later than this pull request.  Cancelling.

ghost:
Actually, I closed this and I should not have.  We should upgrade jersey to 1.19 per this pull request.  I have a touch of dyslexia which led me to temporarily misread the version.

spmallette:
i haven't looked, but what is jersey used for in titan now that Rexster is gone - is there some other component that requires it?

spmallette:
answered my own question, looks like jersey is all up in hadoop/hbase stuff.

ghost:
But you know...when doing all of the SparkGraphComputer testing,  I ended up needing to back jersey down to 1.9 to get stuff to work.  It was like Spark was using an older version and was giving servlet class not found errors, if I recall.  So I am cancelling this particular one after doing that testing.

spmallette:
I think this is main reason I've been a little hesitant on some of your other PRs around version bumps.  The dependencies are a pretty delicate house of cards and the test cases may not catch a big problem on what seems like an innocuous change in version.

PR #1213, opened 2015-12-12
update io-netty version

ghost:
update io-netty version


spmallette:
@dalaro would be nice to bring in this PR to get Titan in line with Gremlin Server at `4.0.29.Final` - not sure what that would affect offhand.  i'm semi-surprised SSL works with Titan/Gremlin Server if that bump isn't in place.

PR #1216, opened 2015-12-12
downgrade guava to prevent hbase errors

ghost:
Guava version 18 causes HBase, which relies on Guava 12 currently, to throw StopWatch exceptions during read/write operations.  Downgrading Guava to version 16 appears to eliminate these errors.  A Jira was opened against HBase to upgrade the version of Guava supported there, but that won't be ready for quite a while.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -780,7 +780,7 @@
                 -->
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>18.0</version>
+                <version>16.0.1</version>
             </dependency>
             <dependency>
                 <groupId>commons-lang</groupId>


dalaro:
I'm not convinced this is safe.  The Guava version is extremely delicate.  We worked around breaking changes to Stopwatch in the past by adding ./titan-hbase-parent/titan-hbase-core/src/main/java/com/google/common/base/Stopwatch.java (which is still in the tree).

Can you share the steps to reproduce the problem, e.g. what test to run and what Maven command?

ghost:
Hi Dan,  Didn't miss this.  Thanks for replying. Will get back to this and reproduce and let you know soon.

ghost:
Hi Dan,  We probably need to move this conversation into a Jira or something and out of here.  I get the sense that you don't want this particular pull request, so I am going to close this and open a Jira to capture the conversation on StopWatch.  Thanks !

Issue #1187, opened 2015-11-16
Conflicting text in bulkloading.txt 


ghost:
The _ID Acquisition Process_ section says `three configuration options that can be tuned to avoid this` but lists only two - `ids.authority.wait-time` and `ids.renew-timeout`. Please fix to update to "two" or list the third. Is it `ids.renew-buffer-percentage`?

dalaro:
I looked at the git history for bulkloading.txt to jog my memory.

#530 removed the third option once listed in this section.  It was `storage.idauthority-retries`.  However, that option was redundant with `ids.renew-timeout` and tended to just make tuning trickier.  Only `ids.renew-timeout` remains today.

The doc section in question talks about avoiding ID block allocation conflicts.  The options with `conflict-avoidance` in their names (listed under http://s3.thinkaurelius.com/docs/titan/1.0.0/titan-config-ref.html#_ids_authority) are relevant.

PR #1226, opened 2015-12-18
Issue1187 fix

ghost:
Based on evidence from the mailing list and docs, this does seem to be the third configuration parameter that was supposed to be documented in the text (three were mentioned, only two were described).  While the default value is .3, I could not uncover a documented "rule of thumb" for this setting during bulk loading so omitted that detail from the next text.

### DIFF
--- a/docs/bulkloading.txt
+++ b/docs/bulkloading.txt
@@ -49,7 +49,6 @@ ID Acquisition Process
 
 When id blocks are frequently allocated by many Titan instances in parallel, allocation conflicts between instances will inevitably arise and slow down the allocation process. In addition, the increased write load due to bulk loading may further slow down the process to the point where Titan considers it failed and throws an exception. There are three configuration options that can be tuned to avoid this.
 
-
 1) `ids.authority.wait-time` configures the time in milliseconds the id pool manager waits for an id block application to be acknowledged by the storage backend. The shorter this time, the more likely it is that an application will fail on a congested storage cluster.
 
 *Rule of thumb*: Set this to the sum of the 95th percentile read and write times measured on the storage backend cluster under load.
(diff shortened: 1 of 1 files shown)


dalaro:
I just commented about this in the associated issue: https://github.com/thinkaurelius/titan/issues/1187#issuecomment-166484940

I do like this paragraph about renew-buffer-percentage, particularly since that option does not seem to be mentioned anywhere in the documentation (according to grep for renew-buff).

However, the link to ID block allocation conflict frequency/avoidance is kind of tenuous.  It does have an indirect effect on conflict frequency, but its direct effect is on write throughput: as the value increases, the chance of Titan blocking as it waits for fresh IDs decreases (up to a point).  Less time spent blocked on fresh IDs means more vertices written per unit time, all else equal, and more frequent ID block allocation attempts/conflicts.  But I'm not sure I would approach it as a conflict tuning parameter, since it's intended to be set at or above the tipping point that results in little or no blocking.

There's a case to be made for mentioning the `conflict-avoidance-*` options listed in http://s3.thinkaurelius.com/docs/titan/1.0.0/titan-config-ref.html#_ids_authority, since the context is about avoiding ID allocation conflicts for better performance.  I might try to write that up and separately present your paragraph on renew-buffer-percentage as a throughput option.

ghost:
Sounds good.  Let me know if now is a good time to cancel my pull request
and I can do that.  I'll wait for your signal.

Here is a build up to another thought on the ID topic in the Titan
documentation:

It is clear that each newly added vertex or edge or property (?) is
assigned a unique id.
Coordinating these IDs between distributed instances of Titan connected to
the same keyspace
or table is clearly necessary. But how this is happens at a more detailed
level is absent and makes some
of the documented tuning parameters a bit mysterious.

Here are a few examples:

The documentation mentions Titan’s id pool manager has a role in acquiring
a set of IDs but it also implies there is something called an ID Authority,
at least in Chapter 12.  What does the use of the word "authority" actually
mean in this case ?

The documentation mentions that all Titan instances must use the same block
size to ensure proper ID allocation but doesn't clarify why.  At first
blush it seems each instance can get any block size, as long as that same
block of IDs isn't handed to some other instance.

The documentation does not explain what goes on during actual "block or ID
generation".  Simplistically, (I realize it isn't this) this could be a set
of sequential numbers in an integer sequence handed back to an instance, so
what takes so long ?

In addition to the changes you are recommending (let me know when you are
done with my pull request text and I can cancel my specific pull request)

I was going to try to dig into the ID block code of Titan and try to figure
some of this out, and provide additional explanations that I think will
clarify some of the numerous settings for this Titan component.

What do you think ?

PR #1230, opened 2015-12-23
auto install spark/giraph into gremlin

ghost:
Spark and Giraph should be auto installed into Titan Gremlin shell.


spmallette:
This was a good idea I think - Noted someone on the list trying to `:install` the plugin because they didn't know it was already bundled.

PR #1231, opened 2015-12-27
Bugfix: handle self-loops (more efficiently)

imri:
As of Titan 1.0, the implementation of TitanVertexDeserializer does not handle scenario of an Edge connecting a Vertex to itself.
In such a case, the same edge loaded twice, once with IN direction, and another time with OUT. Therefore, the edge is added to the graph twice, and while the first addition succeeds, the second causes an 'Edge with id already exists' exception (because there is an attempt to add an existing key to 'graph.edges' dictionary).

As of Titan 1.1, a new method called isLoopAdded was added in order to fix this issue, which loops through the adjacent vertices of a vertex to check if its on a self-loop.

The same purpose can be achieved if you only add the edge once - 
If the edge otherVertexId == vertexId, and the direction is OUT, the vertex is added to the graph, and if the direction is IN, it is ignored.
Doing so prevents the exception.


spmallette:
This pull request seems to come with a lot of "extra commits" or something. it makes it a bit hard to evaluate.  could you please resubmit this PR against the titan11 branch and we can evaluate it at that point?

analytically:
+1

technotrance:
Call green lol

PR #1247, opened 2016-01-21
fix compat loader value for hbase

ghost:
The DEFAULT_HBASE_CLASS_NAME value in HBaseCompatLoader specifies HBaseCompat1_00.  Running on an HBase 1.1.1 cluster, the following error happens when running SparkGraphComputer with that value:

```
Caused by: java.lang.ClassNotFoundException: com.thinkaurelius.titan.diskstorage.hbase.HBaseCompat1_00
    at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    at java.lang.Class.forName0(Native Method)
    at java.lang.Class.forName(Class.java:264)
    at com.thinkaurelius.titan.diskstorage.hbase.HBaseCompatLoader.getCompat(HBaseCompatLoader.java:54)
```

To work around this problem, this property:

```
storage.hbase.compat-class=com.thinkaurelius.titan.diskstorage.hbase.HBaseCompat1_0
```

can be set in the hadoop-conf hbase file.

Looking at line 37, where dots are replaced with _,  1.0 will end up being 1_0 not 1_00.
When DEFAULT_HBASE_CLASS_NAME is changed to HBaseCompat1_0, then
the override property in the properties file is not needed and the code works with HBase 1.1.1.

### DIFF
--- a/titan-hbase-parent/titan-hbase-core/src/main/java/com/thinkaurelius/titan/diskstorage/hbase/HBaseCompatLoader.java
+++ b/titan-hbase-parent/titan-hbase-core/src/main/java/com/thinkaurelius/titan/diskstorage/hbase/HBaseCompatLoader.java
@@ -12,7 +12,7 @@ public class HBaseCompatLoader {
 
     private static final String DEFAULT_HBASE_COMPAT_VERSION = "1.0";
 
-    private static final String DEFAULT_HBASE_CLASS_NAME = "com.thinkaurelius.titan.diskstorage.hbase.HBaseCompat1_00";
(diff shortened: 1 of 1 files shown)


kottmann:
I have the same problem. Running titan against HBase 1.1.x throws the error described above and I would recommend accepting the pull request to correct the class name.

Issue #1254, opened 2016-01-29
Docs, Chapter 3, select() fails


leifurhauks:
Following the docs, section 3.3.2.2 Pluto's Brothers, the following query fails:

```
gremlin> g.V(pluto).out('brother').as('god').out('lives').as('place').select()
Could not find which method select() to invoke from this list:
  public org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal#select(java.lang.String)
  public org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal#select(org.apache.tinkerpop.gremlin.process.traversal.Pop, java.lang.String)
  public transient org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal#select(java.lang.String, java.lang.String, [Ljava.lang.String;)
  public transient org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal#select(org.apache.tinkerpop.gremlin.process.traversal.Pop, java.lang.String, java.lang.String, [Ljava.lang.String;)
```

So the method `select` without any parameters apparently does not exist?

The following does work:

```
gremlin> g.V(pluto).out('brother').as('god').out('lives').as('place').select('god', 'place')
==>[god:v[8320], place:v[4224]]
==>[god:v[8424], place:v[4328]]
```

This example in the docs should be updated to reflect the traversal API.

spmallette:
yeah - i think you have to specify the labels explicitly now.  don't remember the reason for that offhand. thanks for mentioning this.  if you'd like to submit a PR i can get that merged - please submit to the titan11 branch if you intend to do that.

ghost:
Hope I didn't miss it, but didn't see a Pull Request for this - I'll submit one and return here with the PR number to close the loop.

spmallette:
PR #1255 was merged.

PR #1255, opened 2016-01-29
revise uses of select() without labels in examples

leifurhauks:
The zero parameter signature of select() no longer exists; this fixes two of the example queries to use explicit labels (re: #1254 )


spmallette:
merged - thanks for fixing that.

PR #1263, opened 2016-02-11
Fix wrong cassandra inputformat configuration

petteraas:
When testing hadoop on titan 1.0.0 I had a problem with that it tried to use the keyspace 'titan', I couldn't at first figure out why because I used a copy of this file, but when I took a closer gander at this file I found that the '.cassandra' part was missing on the keyspace line.


pluradj:
+1 you're right

spmallette:
Good find - thanks for the PR

Issue #1242, opened 2016-01-14
gremlin.bat and gremlin-server.bat issues in titan-1.0.0


pluradj:
An old thread, but still relevant in titan10 and titan11 branches:
https://groups.google.com/d/msg/aureliusgraphs/tZ0XhwJY9jQ/R-jboJBWAgAJ

More recently
http://stackoverflow.com/questions/34793406/activate-tinkerpop-plugins-for-titan

If you unzip titan-1.0.0-hadoop1.zip on Windows, you'll see errors out of the box.

```
C:\titan-1.0.0-hadoop1>bin\gremlin.bat
Error opening zip file or JAR manifest missing : ..\lib\jamm-0.3.0.jar
Error occurred during initialization of VM
agent library failed to init: instrument

C:\titan-1.0.0-hadoop1\bin>gremlin.bat
The input line is too long.
The syntax of the command is incorrect.
```

I think the fix is to get the gremlin*.bat files better aligned with the latest from TinkerPop3. There are some additional Titan-specific changes required for the jamm-0.3.0.jar and the slf4j/logback conflict.

For the gremlin-server

```
C:\titan-1.0.0-hadoop1>bin\gremlin-server.bat
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/C:/titan-1.0.0-hadoop1/lib/logback-classic-1.1.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/C:/titan-1.0.0-hadoop1/lib/slf4j-log4j12-1.7.5.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
16:19:38.362 [main] DEBUG i.n.u.i.l.InternalLoggerFactory - Using SLF4J as the default logging framework
16:19:38.378 [main] INFO  o.a.t.gremlin.server.GremlinServer -
         \,,,/
         (o o)
-----oOOo-(3)-oOOo-----

16:19:38.378 [main] ERROR o.a.t.gremlin.server.GremlinServer - Configuration file at conf/gremlin-server.yaml could not be found or parsed properly. [conf\gremlin-server.yaml (The system cannot find the file specified)]
```

[...]

PR #1274, opened 2016-03-04
fixed Windows scripts based on TinkerPop

pluradj:
https://github.com/thinkaurelius/titan/issues/1242

Fixes for `gremlin.bat` and `gremlin-server.bat` based off of the TinkerPop3 versions of those files.
- I tested on Windows 7 and Windows 10 with Cassandra 2.1.13 and Elasticsearch 1.5.2 local on the Windows machine
- I tested OLAP traversals (Gryo-based) against a remote Linux-based HDFS/YARN/Spark cluster
- I tested passing scripts in with `gremlin.bat -e`
- I tested installing plugins with `gremlin-server.bat -i`
- I did not test OLAP against a Windows HDFS/YARN/Spark cluster. It became too much trouble to install -- do people really run Hadoop on Windows :)

I included a `logback.xml` more as an example, but it doesn't get used since the scripts force log4j ahead of it on the classpath.

It would be simpler to pull in the `winutils.exe` directly into the build, but I imagine that could complicate licensing concerns.

I didn't tackle a conversion of `titan.sh` to `titan.bat` with this PR. I'll open up a separate issue for that. Another separate issue would be adding more Titan-specific support for `gremlin.bat -v`.

### DIFF
--- a/titan-core/src/main/java/com/thinkaurelius/titan/core/Titan.java
+++ b/titan-core/src/main/java/com/thinkaurelius/titan/core/Titan.java
@@ -2,6 +2,8 @@ package com.thinkaurelius.titan.core;
 
 import com.thinkaurelius.titan.graphdb.configuration.TitanConstants;
 
+import org.apache.tinkerpop.gremlin.util.Gremlin;
+
 
 /**
  * Contains constants for this Titan Graph Database.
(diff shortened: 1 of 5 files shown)


spmallette:
Nice PR. Merging - thanks.

PR #1280, opened 2016-03-08
code examples for full-text search and string search

pluradj:
http://stackoverflow.com/questions/35835681/tinkerpop-api-does-not-use-mixed-elasticsearch-index-while-retrieving-data-from


spmallette:
thanks @pluradj

Issue #1139, opened 2015-09-03
Shade Guava


dalaro:
We upgraded to Guava 18 in #854, but we didn't shade Titan's copy of Guava.

We should do that to give users one less Guava version constraint to worry about.

ghost:
@dalaro  This seems to be a blocker for SparkGraphComputer to work with HBase, so I am looking at this.

PR #1282, opened 2016-03-10
id: 1278 Pin guava version to 18

fcalderonnearsoft:
desc: added <version>18.0</version> to pom.xml
 reviewer: sayrer

### DIFF
--- a/titan-core/pom.xml
+++ b/titan-core/pom.xml
@@ -85,6 +85,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
+            <version>18.0</version>
         </dependency>
         <dependency>
             <groupId>com.carrotsearch</groupId>


fcalderonnearsoft:
@spmallette @dalaro is there any chance that someone could check this, please?

spmallette:
ha - that's going to have to be a @dalaro job. did you run the full test suite successfully?

fcalderonnearsoft:
@spmallette Yes, I did! :)

okram:
We shouldn't "pin" anything. There will always be version differences between different users. Instead either Titan uses shading or the user shades their jars.

spmallette:
yeah - looks like @dalaro had shading in mind: #1139

fcalderonnearsoft:
@spmallette @okram What do you mean with 'shading' ?

spmallette:
It basically would repackage guava for a particular version under the titan distribution. You can see how it's done here in TinkerPop for `gremlin-shaded`:

https://github.com/apache/incubator-tinkerpop/blob/49ed3612c54c4e056796d654c24de6bb9496e775/gremlin-shaded/pom.xml

Looking at that, you can see that we repackaged kryo, for example, with a `org.apache.tinkerpop.shaded.kryo` package name. That's at least one use case of shading anyway.  You should probably read up on the Maven shade plugin to learn more.

PR #1141, opened 2015-09-04
make the number of vertices in testVertexCentricQuery configurable

amcp:

PR #1291, opened 2016-04-04
make noVertices configurable

amcp:
old PR https://github.com/thinkaurelius/titan/pull/1141
I accepted the CLA - should be OK to merge


titan-test/src/main/java/com/thinkaurelius/titan/graphdb/TitanGraphTest.java
```diff
@@ -2518,8 +2518,12 @@ private void awaitAllThreadsReady() {
      ==================================================================================*/
 
     @Test
-    @SuppressWarnings("deprecation")
     public void testVertexCentricQuery() {
+        int noVertices = 10000;
```
spmallette: is this what you wanted to do here? this is different from the previous PR that called `testVertexCentricQuery(10000)`

amcp:
fixing

amcp:
fixed

PR #1295, opened 2016-04-04
Allow custom vertex id in TitanBlueprintsTransaction when graph.set-vertex-id=true

amcp:
The following TitanVertexFeature suggests that the use of custom vertex ids through TinkerPop Structure depends on the allowVertexIdSetting() value:
https://github.com/thinkaurelius/titan/blob/titan11/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/tinkerpop/TitanFeatures.java#L173

However the TitanBlueprintsTransaction implementation forbids custom ids regardless of the allowVertexIdSetting():
https://github.com/thinkaurelius/titan/blob/titan11/titan-core/src/main/java/com/thinkaurelius/titan/graphdb/tinkerpop/TitanBlueprintsTransaction.java#L90

Making me drill down to vendor-specific implementation to use custom vertex ids even though they should be supported, like this:
https://github.com/classmethod/tupl-titan-storage-backend/blob/titan10/src/test/java/jp/classmethod/titan/TuplStorageSetup.java#L100

Even though I should be able to use graph.addVertex(T.id, titanVertexId, T.label, label, NODE_ID, longVal) because I have enabled setting vertex IDs:
https://github.com/classmethod/tupl-titan-storage-backend/blob/titan10/src/test/java/jp/classmethod/titan/TuplStorageSetup.java#L124


amcp:
@dalaro I think this is a bug in Titan...

PR #1310, opened 2016-04-27
Fix length check for sliceStart

biyuhao:


biyuhao:
I think this is a minor fix, would you mind have a look at this please? @spmallette

PR #1313, opened 2016-05-04
fix gremlin server example config file error

boliza:


spmallette:
First, these aren't gremlin server configuration files, they are configuration files for the driver. Second, they look ok to me. See a samples from the TinkerPop repository:

https://github.com/apache/incubator-tinkerpop/blob/3.0.2-incubating/gremlin-console/conf/remote.yaml

https://github.com/apache/incubator-tinkerpop/blob/3.0.2-incubating/gremlin-console/conf/remote-objects.yaml

Thoughts?

PR #1345, opened 2016-09-06
Fix ConsistentKeyIDAuthority format string

srosenthal:
Previously the formatter %d was used for a java.time.Duration. When this code gets executed, instead of throwing the TemporaryLockingException, it would instead throw an IllegalFormatConversionException with the message "d != java.time.Duration".


BillSobel:
I believe this was fixed in 1174

PR #1376, opened 2017-02-18
Update readme to add notice for JanusGraph project

manigandham:
With TitanDB effectively stalled, it would be best to redirect attention to the new JanusGraph project so that we can continue the interest and efforts of users without confusion.


laughedelic:
👍

manigandham:
@pluradj @spmallette any help/interest in putting this notice in place?

pluradj:
I do not have any permission or authority on the thinkaurelius project to do this.

